PR #2, opened 2020-09-16
Fixes/update readme

FionaDL:
This PR adds an updated README to the points application. 

**What is this PR:**

- [ ] Bug fix
- [ ] Feature
- [X] Chore


manuca (approved):
Looks great.

FionaDL:
@rdormer I guess since this can't really deploy to staging the next step is a merge?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at ernesto+github@ombulabs.com. All
```
etagwerker, line 58: @FionaDL Let's use oss@ombulabs.com here.

README.md
```diff
@@ -7,7 +7,7 @@ This is a Rails application to collaboratively estimate stories
 To get started with the app, clone the repo and then install the needed gems:
 
 ```
-$ bundle install 
+$ bundle install
```
etagwerker, line 10: @FionaDL I think `./bin/setup` is more comprehensive: https://github.com/fastruby/points/blob/main/bin/setup 

Maybe we should instruct people to run that instead?

etagwerker (commented):
@FionaDL Looks good. I added a couple of comments. 💯

README.md
manuca: We talked a bit about the bin/setup with Fiona. I mentioned that being such a standard Rails setup maybe it would be less surprising for a new comer to use the usual steps.  

Personally I'm more in favor of being explicit in these cases and users will still need to make tweaks to the `.sample` files, etc so the setup won't buy them much.

FionaDL: @etagwerker @manuca where did we land on this?

manuca: I'm fine either way, just wanted to give my opinion about it but its not a strong one at all.
BTW I haven't tried running the setup, just looked at it. Did we test if it's working?

etagwerker: @manuca @FionaDL I think we should go with `./bin/setup` 

New Rails applications come with a `bin/setup` script. So the usual steps when I start working on a Rails app includes running `./bin/setup` -- This is what the template looks like: https://github.com/rails/rails/blob/bd49578cf5c8d181950a172e231c9886f54bbd58/railties/lib/rails/generators/rails/app/templates/bin/setup.tt

manuca: Interesting, didn't know about this fact. Sorry for the confusion I might have added to the discussion then.

etagwerker: @manuca Cool, no problem. I'm glad we are all on the same page now. 😄

Issue #16, opened 2021-01-25
Add support for subprojects


arielj:
When we have to do a big upgrade involving different stages and versions jumps, it would be nice to have each thing as a project and one project containing all.

For example: let's say we want to upgrade an app from rails 3.0 to rails 5.0, but also do some big changes in the middle like replacing a gem with an alternative version that should be handled outside the usual upgrade project.

We would have multiple projects:
- Rails 3.0 to 3.1
- Rails 3.1 to 3.2
- Ruby 1.9.3 to 2.3
- Replace X gem with Y replacement
- Rails 3.2 to 4.0
- Rails 4.1 to 4.2
- Rails 4.2 to 5.0

And then a `parent` project to group the rest.

That way, each stage/story of a project can be estimated as a unit and then the parent project can generate the report.

Currently, adding all those projects makes the index full of similar entries and having all the stories into one single project makes it hard to organize them and it can't show the estimation for each separately.

rdormer:
@bronzdoc @arielj so I'm thinking the best way to go about this would be to introduce some type of higher level parent object that can own more than one project.  Then when you create a project, by default you're creating the higher level object with a single child project, and that can be handled in the UI as a special case.  If you create more projects for that parent object, then we'd need to make UI changes to accomodate showing the projects specific to that parent.

Does the basic approach seem sound?  I think it'll be a lot clearer and easier than trying to have a project that `has_many` of itself....

arielj:
While writing this issue I imagined something like:

project belongs_to parent, class_name: 'Project'
project has_name projects

Then in the index we do something like `@projects = Project.where(parent_id: nil)` to list only the projects that are not sub projects. That way we don't need to create higher level projects for the projects we already have in the index, and also we don't need to change anything for the Story model (if we have 2 different kinds of project we would need to make the Story > project some polymorphic association).

Then in the show action, if the project has subprojects we can list them (and then maybe we can have a list of stories for this parent project too), something like:

```
This Super Large Project

Subprojects:
- part 1
- part 2

Stories:
- some story for the parent
- another story for the parent

actions: add story, add subproject, edit, etc
```

If that makes sense I prefer that instead of adding another class

rdormer:
@bronzdoc @arielj do you think it makes sense to include reporting in this feature?  I'm inclined to push that off to another separate story.  At the very least, we need to define how we want to handle compiling and exporting reports for projects with subprojects.

arielj:
[...]

PR #18, opened 2021-02-03
Add subprojects

rdormer:
**Description:**

This PR adds support for projects nested under other projects (sub-projects).  The nested project is a child of another project, and is displayed and edited accordingly.  See #16 for more  details.
Screen Shot 2021-02-03 at 1 05 36 PM

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -28,7 +28,7 @@ class ProjectsController < ApplicationController
       redirect_to "/projects"
     else
       flash[:error] = @project.errors.full_messages
-      render "projects/new"
+      redirect_back(fallback_location: "projects/new")
     end
   end
 
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -56,9 +56,14 @@ class ProjectsController < ApplicationController
     end
   end
 
+  def new_sub_project
+    @project = Project.find(params[:project_id])
+    @sub = Project.new(parent_id: @project)
+  end
+
   private
 
   def projects_params
-    params.require(:project).permit(:title, :status)
+    params.require(:project).permit(:title, :status, :parent_id)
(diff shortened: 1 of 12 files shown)


bronzdoc (approved):

bronzdoc:
@arielj Can you QA this?

bronzdoc:
Closes https://github.com/fastruby/points/issues/16

arielj:
Doing the QA now, adding comments while testing it:

- This may be too subjective, but the `add subproject` is hard to find unless you know where to look for it, right now we need to go to edit a project to be able to add sub projects and it's not a screen you would think of going since the create form only has a title. Maybe it's more clear to display that option in the project's `show` page. Maybe next to the `edit or delete project` button?

- It's hard to navigate between the parent project and the subprojects, to go from parent to a child I have to go to the index and click the child, then to go back I have to go back again to the index. Maybe we can add a link to the parent project next to a child's project name in its `show` page, and add a list of subprojects right next to the parent's title (or maybe as a sidebar kind of thing?). We could have the `add subproject` action close to the list of subprojects too, like moving what we have on the edit page to the show page would make things faster to navigate

- When I try to create a subproject with no title, I have no feedback of the error (if I reload the page or move to another page I see the flash message), this is probably an issue with form_with doing an ajax request by default

- When editing a sub project, there's a double line, I guess moving the `hr` tag inside the `if @project.projects.count > 0` is enough:

image

@bronzdoc @rdormer here's the QA

arielj:
@fastruby/black-bunny-brigade one more QA thing (most of the previous feedback is fixed 👍 )

When you go to `Add subproject`, clicking `Back` takes you to the index instead of the previous page. It should take you to the parent's show page again.

bronzdoc (approved):
This looks good @rdormer I think just one more comment to address from Ariel and we are good to go

PR #43, opened 2021-05-26
Renames 'Delete' to 'Bulk Delete' and adds visual cues

torresga:
**Description:**

This PR _Closes #30_.

<s>I decided to do two things for this feature: 

1. I renamed the magenta ‘Delete’ button to “Bulk Delete”, because I found it confusing to have two buttons named ‘Delete’.
2. I did some research about disabling the ‘Bulk Delete’ button when no stories are selected, but after reading that disabled buttons are bad for accessibility (particularly https://axesslab.com/disabled-buttons-suck/), I decided to keep the button enabled, and instead added an alert that tells the user to select one or more stories. You can see a video below:</s>

As discussed below, I renamed the magenta ‘Delete Button’ to ‘Bulk Delete’, and made it lighter/disabled it until 1 or more stories are selected. When 1 or more stories are selected, the number of stories are included in the button text.

## Before:
Initial state:
Screen Shot 2021-06-02 at 12 12 28 PM
When a story is selected:
Screen Shot 2021-06-02 at 12 12 42 PM

## After:
Initial state:
Screen Shot 2021-06-02 at 12 07 41 PM

When a story is selected:
Screen Shot 2021-06-02 at 12 08 20 PM

The tests are currently failing due to the Mimemagic 0.3.5 dependency, which is updated in this commit: _6e5eb49_. 

Please let me know if you have any suggestions.

### DIFF
--- a/app/assets/javascripts/project.js
+++ b/app/assets/javascripts/project.js
@@ -1,4 +1,4 @@
-$(() => {
+document.addEventListener("turbolinks:load", function() {
   $("input[name='stories[]']").click(() => {
     let selected = $("input[name='stories[]']:checked");
 
--- a/app/assets/javascripts/project.js
+++ b/app/assets/javascripts/project.js
@@ -1,10 +1,17 @@
 $(() => {
   $("input[name='stories[]']").click(() => {
     let selected = $("input[name='stories[]']:checked");
-    let ending = selected.length == 1 ? "y" : "ies";
-    $(".number-of-stories-selected")
-      .text(`${selected.length} stor${ending} selected`)
-      .removeClass("alert");
+
+    if (selected.length > 0) {
+      let ending = selected.length == 1 ? "y" : "ies";
+      $("#bulk_delete")
+        .text(`Bulk Delete (${selected.length} Stor${ending})`)
+        .removeAttr("aria-disabled");
+    } else {
+      $("#bulk_delete")
+        .text(`Bulk Delete`)
+        .attr("aria-disabled", "true");
+    }
   })
 
   $("#bulk_delete").click(() => {
@@ -12,12 +19,6 @@ $(() => {
     $("input[name='stories[]']:checked").each((_, checkbox) => {
       stories_ids.push($(checkbox).val())
     })
-
-    if (stories_ids.length === 0) {
-      $(".number-of-stories-selected")
-        .text("Please select 1 or more stories")
-        .addClass("alert");
-    }
 
     let token = $("meta[name='csrf-token']").attr("content")
     $.ajaxSetup({
@@ -33,7 +34,6 @@ $(() => {
       success: () => {
         $(stories_ids).each((_, id) => {
           console.log(id)
-          $(".number-of-stories-selected").text("");
           $("#story_" + id).remove();
         })
       },
--- a/app/assets/javascripts/project.js
+++ b/app/assets/javascripts/project.js
@@ -20,12 +20,6 @@ $(() => {
       stories_ids.push($(checkbox).val())
     })
 
-    if (stories_ids.length === 0) {
-      $(".number-of-stories-selected")
-        .text("Please select 1 or more stories")
-        .addClass("alert");
-    }
-
     let token = $("meta[name='csrf-token']").attr("content")
     $.ajaxSetup({
       beforeSend: function (xhr) {
@@ -40,7 +34,6 @@ $(() => {
       success: () => {
         $(stories_ids).each((_, id) => {
           console.log(id)
-          $(".number-of-stories-selected").text("");
           $("#story_" + id).remove();
         })
       },
--- a/app/assets/javascripts/project.js
+++ b/app/assets/javascripts/project.js
@@ -1,9 +1,9 @@
 document.addEventListener("turbolinks:load", function() {
   $("input[name='stories[]']").click(() => {
-    let selected = $("input[name='stories[]']:checked");
+    const selected = $("input[name='stories[]']:checked");
 
     if (selected.length > 0) {
-      let ending = selected.length == 1 ? "y" : "ies";
+      const ending = selected.length == 1 ? "y" : "ies";
       $("#bulk_delete")
         .text(`Bulk Delete (${selected.length} Stor${ending})`)
         .removeAttr("aria-disabled");
--- a/app/assets/javascripts/project.js
+++ b/app/assets/javascripts/project.js
@@ -6,11 +6,13 @@ document.addEventListener("turbolinks:load", function() {
       const ending = selected.length == 1 ? "y" : "ies";
       $("#bulk_delete")
         .text(`Bulk Delete (${selected.length} Stor${ending})`)
-        .removeAttr("aria-disabled");
+        .attr("aria-disabled", "false")
+        .prop("disabled", false);
     } else {
       $("#bulk_delete")
         .text("Bulk Delete")
-        .attr("aria-disabled", "true");
+        .attr("aria-disabled", "true")
+        .prop("disabled", true)
     }
   })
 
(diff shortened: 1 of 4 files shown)


buys-fran:
Hey @torresga , good for taking a different approach on this one.

I would like to present a different approach as I don't fully agree with https://axesslab.com/disabled-buttons-suck/.
I realise that contrast is important but according to the official web content accessibility guidelines, it is okay for deactivated components to have lower contrast. https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

We can disable the button, and and only reduce its opacity (not fully grey it out).
We can also use aria-disabled for users that make use of screen readers https://a11y-101.com/development/aria-disabled

Finally if none of these seem like good option, perhaps we can only show the button when stories are selected, and set is to display: none when nothing is selected.

The alert makes the UI seem jumpy since it moves the entire screen down. It is also green(ish), which seems misleading as green us usually associated with successful actions.

Let me know what you think?

torresga:
> Hey @torresga , good for taking a different approach on this one.
> 
> I would like to present a different approach as I don't fully agree with https://axesslab.com/disabled-buttons-suck/.
> I realise that contrast is important but according to the official web content accessibility guidelines, it is okay for deactivated components to have lower contrast. https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html
> 
> We can disable the button, and and only reduce its opacity (not fully grey it out).
> We can also use aria-disabled for users that make use of screen readers https://a11y-101.com/development/aria-disabled
> 
> Finally if none of these seem like good option, perhaps we can only show the button when stories are selected, and set is to display: none when nothing is selected.
> 
> The alert makes the UI seem jumpy since it moves the entire screen down. It is also green(ish), which seems misleading as green us usually associated with successful actions.
> 
> Let me know what you think?

Hi @buys-fran ,

Thank you for taking the time to review my PR. After thinking over what you said, here is what I propose:

1. For the 'Bulk Delete' button, I will decrease the opacity to 0.5, which will hopefully be a visual cue that the button is disabled. I will add `aria-disabled="true"` to that button. So I am thinking it will look something like this: 
Screen Shot 2021-05-27 at 3 15 58 PM

2. For the second part, I was thinking of adding the # of stories selected to the text when the checkboxes are selected. That way we can get rid of the alert that moves everything else on the page down. Something like this: 
Screen Shot 2021-05-27 at 3 15 23 PM

What do you think? I agree that the alerts shouldn't be green. However, I noticed that all of the alerts are currently green - regardless of whether they are success or error messages. Maybe we should resolve that in another PR?

buys-fran:
@torresga I think the use of opacity looks really good. Nice 🎉

I think adding a n issue for the alerts is a good idea 👍  Perhaps red for errors and green for notification.

I would only see what it looks like if you changed the button text to:
`Bulk Delete (3)` - I think the number is enough to indicate that 3 stories will be deleted.
or
`Bulk Delete (3 Stories)` could also work if the concensus is that `(3)` on its own is not clear enough.

torresga:
https://user-images.githubusercontent.com/6892410/120046758-df029500-bfe0-11eb-9aea-1f256459d986.mov

@buys-fran What do you think? I added (# stories) just because I like to err on the side of being super clear.

buys-fran:
>  bulk-delete-updated.mov 
> @buys-fran What do you think? I added (# stories) just because I like to err on the side of being super clear.

I think it looks good! Nice one! 🚀

Could you please add a picture of the before and after UI in the PR description?
I think it would make it quick to see exactly what changed.

buys-fran:
I tested the new button locally and it works as expected. ✅

@torresga Could you also please add a note to your description explaining that the tests are failing due to the Mimemagic 0.3.5 dependency. The dependency is updated in this commit: 6e5eb49d33012c60c258dea8290ed7704641e3e4

torresga:
@buys-fran Thank you for the feedback! I updated my description.

bronzdoc:
just rebase `main` here as well to make the build pass

buys-fran:
@torresga are you comfortable to rebase - or would you like to pair so we can do it together?

torresga:
@buys-fran @bronzdoc Ok, I rebased `main` into this feature to make the build pass.

buys-fran:
I tested this buttun in chrome and safar, both works great! 
Nice one @torresga 🎉

torresga:
@mateusdeap Can you please review this PR for me? It's been QAed by @buys-fran and myself 😸

torresga:
I think you already partially reviewed this @arielj but I wanted to get another pair of eyes on this (so hopefully we can merge soon 😸 )

app/assets/javascripts/project.js
```diff
@@ -1,10 +1,29 @@
-$(() => {
-  $("#bulk_delete").click(() => {
+document.addEventListener("turbolinks:load", function() {
+  $("input[name='stories[]']").click(() => {
+    let selected = $("input[name='stories[]']:checked");
```
arielj, line 3: this can be `const selected = ...`

app/assets/javascripts/project.js
```diff
@@ -1,10 +1,29 @@
...
+
+    if (selected.length > 0) {
+      let ending = selected.length == 1 ? "y" : "ies";
```
arielj, line 6: this can be `const ending = ...`

arielj (approved):
This looks good, I just added 1 small suggestion to use `const` instead of `let` but it's all good.

torresga:
Ok, I put the "Ready to QA" label on this because I am not sure where to go from here. @buys-fran and I have already QA'ed this locally. Is the next step to QA on staging?

buys-fran:
I was testing it out one last time before merging, and noticed that it still submits a requests to the server when the button is disabled.

To QA. run `rails s` and have the server logs open. Click on the disabled bulk delete button and notice that the logs show a delete request.

buys-fran:
@torresga , I meant to tag you in the previous comment 🙂

torresga:
@buys-fran Ok, I will take a look at this

torresga:
@buys-fran So the problem with `aria-disabled` is that it doesn’t actually disable the button…it avoids the accessibility problems that `disabled` has (the screen reader will get the fact that there is a button and it will tell the user that the button is dimmed or disabled), but as you have seen, the button can still submit a request to the server. If this is a problem, maybe we can add the `disabled ` attribute as well as the `aria-disabled` attribute (though I don't know the implications of this.) Or try combining `aria-disabled` with `aria-live` to prevent the user from making a request to the server by accident as mentioned here: https://a11y-101.com/development/aria-disabled

buys-fran:
> @buys-fran So the problem with `aria-disabled` is that it doesn’t actually disable the button…it avoids the accessibility problems that `disabled` has (the screen reader will get the fact that there is a button and it will tell the user that the button is dimmed or disabled), but as you have seen, the button can still submit a request to the server. If this is a problem, maybe we can add the `disabled ` attribute as well as the `aria-disabled` attribute (though I don't know the implications of this.) Or try combining `aria-disabled` with `aria-live` to prevent the user from making a request to the server by accident as mentioned here: https://a11y-101.com/development/aria-disabled

Apologies for my delayed response. I meant to reply earlier. ⏰
I think we should keep the aria-disabled. I am not sure we should add aria-live as it is not technically a single form with one submit button. So multiple indications 📣📣 that a button has become live might be overwhelming.

The way I see it, we have 2 options.
1. ✅ Use both aria-disabled + disabled
2. ✅ Use aria-disabled + `event.preventDefault();`

I suspect 1 might be easier for now. What do you think? 🤷

torresga:
@buys-fran Yeah I just did Option 1 (aria-disabled + disabled) for now, we can always revisit..

etagwerker:
@torresga @buys-fran I just tested this and it looks good. Thanks!

Issue #20, opened 2021-02-11
[Feature Request] Ability to archive projects


arielj:
It would be nice to be able to archive older projects. Right now the index for Points is full of projects finished long ago and the new and active ones are at the bottom.

Maybe we would need some checkbox or something in the index action to show/hide (hidden by default) the archived projects just in case.

FionaDL:
I am currently working on this.

PR #44, opened 2021-05-26
Add feature to archive projects. 

FionaDL:
This PR deals with issue #20 BY @arielj 
I added the ability to archive and unarchive specific projects, as well as to show or hide archived projects on the index page. I also added two tests. One thing that I didn't add, but might be a good idea to include on a future story is a button to show all projects at the same time, archive and unarchived. 

Screen Shot 2021-05-26 at 12 00 00 PM
Screen Shot 2021-05-26 at 12 00 12 PM
Screen Shot 2021-05-26 at 12 00 22 PM
Screen Shot 2021-05-26 at 12 00 30 PM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -161,7 +161,9 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (1.0.0)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
+      nokogiri (~> 1)
+      rake
     mini_mime (1.0.2)
     mini_portile2 (2.5.0)
     minitest (5.14.3)
@@ -367,6 +369,7 @@ DEPENDENCIES
   jquery-rails
   jquery-ui-rails (~> 5.0, >= 5.0.5)
   listen (>= 3.0.5, < 3.2)
+  mimemagic (~> 0.3.8)
   next_rails
   omniauth-github
   omniauth-rails_csrf_protection
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -22,8 +22,8 @@ class ProjectsController < ApplicationController
     head :ok
   end
 
-  def toggle_archived
-    Project.find(params[:id]).toggle_archived! 
+  def toggle_archive
+    Project.find(params[:id]).toggle_archived!
   end
 
   def create
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -2,11 +2,8 @@ class ProjectsController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    if params[:archived] == "true"
-      @projects = Project.where(parent_id: nil, status: "archived")
-    else
-      @projects = Project.where(parent_id: nil, status: nil)
-    end
+    status = params[:archived] == "true" ? "archived" : nil
+    @projects = Project.where(parent_id: nil, status: status)
   end
 
   def new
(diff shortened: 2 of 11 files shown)


Gemfile.lock
```diff
@@ -161,7 +161,9 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (1.0.0)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
```
arielj, line 164: can you add also the Gemfile.next.lock with this change?

app/controllers/projects_controller.rb
```diff
@@ -2,7 +2,11 @@ class ProjectsController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @projects = Project.where(parent_id: nil)
+    if params[:archived] == "true"
+      @projects = Project.where(parent_id: nil, status: "archived")
+    else
+      @projects = Project.where(parent_id: nil, status: nil)
+    end
```
arielj, line 9: What about doing something like:

```rb
status = params[:archived] == "true" ? "archived" : nil
@projects = Project.where(parent_id: nil, status: status)
```

app/controllers/projects_controller.rb
```diff
@@ -21,6 +25,15 @@ def sort
     head :ok
   end
 
+  def toggle_archive
+    @project = Project.find(params[:project_id])
+    if !@project.archived?
+      @project.update_column :status, "archived"
+    else
+      @project.update_column :status, nil
+    end
```
arielj, line 34: What do you thinkg about something like:

```rb
def toggle_archived
  Project.find(params[:id]).toggle_archived! # check this should be params[:id], not params[:project_id]
end
```

and then in the model something like:

```rb
def toggle_archived!
  new_status = archived? ? nil : "archived"
  update_column :status, new_status
end
```

app/views/projects/show.html.erb
```diff
@@ -64,7 +64,7 @@
     <% unless @project.parent_id %>
       <%= link_to 'Add Sub-Project', project_new_sub_project_path(@project.id), id:"subproject", class: "button magenta" %>
     <% end %>
-
+    <%= link_to "Archive Project", project_toggle_archive_path(@project.id), method: "patch", id: "archive", class: "button magenta", remote: true %>
```
arielj, line 67: maybe this id can be `toggle_archive` instead of `archive` since it does both things?

spec/features/projects_manage_spec.rb
```diff
@@ -29,6 +30,12 @@
     expect(page).to have_content "Project updated!"
   end
 
+  it "allows me to archive a project", js: true do
+    visit project_path(id: project.id)
+    click_link "Archive Project"
+    expect(page).to have_content "Unarchive Project"
```
arielj, line 36: you should also add some check for the actual project like

```rb
expect(project.reload).to be_archived
```

arielj (commented):
Hey @FionaDL it looks good, I added some refactor suggestions and an improvement in the tests, I think with that extra expectation on the tests they are enough the test this new feature

app/controllers/projects_controller.rb
```diff
@@ -21,6 +22,10 @@ def sort
     head :ok
   end
 
+  def toggle_archived
```
arielj, line 25: Sorry, in my suggestion I said `toggle_archived` here but the route is actually `toggle_archive`

arielj (approved):
This is good, I'm not sure who should to the QA

@fastruby/black-bunny-brigade

PR #45, opened 2021-05-26
Updates files with standardrb linter suggestions

torresga:
**Description:**

This PR fixes all the problems that `standardrb` found.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -1,10 +1,10 @@
-require 'csv'
+require "csv"
 class StoriesController < ApplicationController
   before_action :authenticate_user!
   before_action :find_project, except: :bulk_destroy
   before_action :find_story, only: [:edit, :update, :destroy, :show]
 
-  CSV_HEADERS = %w{id title description position}
+  CSV_HEADERS = %w[id title description position]
 
   def new
     @story = Story.where(id: params[:story_id]).first_or_initialize.dup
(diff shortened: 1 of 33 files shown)


app/controllers/stories_controller.rb
```diff
@@ -56,36 +56,39 @@ def update
   end
 
   def import
-
     if !params[:file].original_filename.ends_with?(".csv")
       flash[:error] = "Invalid File: Must be CSV"
-      redirect_to(@project) and return
+      redirect_to(@project) && return
```
arielj, line 61: hey @gelsey, everything looks good, I'm just wondering if these changes from `and` to `&&` work as expected because the documentation states that we should use `and return` https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors.

I think the problem is prevented because we have `redirect_to(@project)` instead of `redirect_to @project`, I'm not sure if there's a quick way to test this manually just to verify.

arielj (approved):
it looks good! I just added one comment that maybe it's worth to verify one code change, but otherwise this is good.

app/controllers/stories_controller.rb
torresga: Hi @arielj,
I think you are right that `redirect_to(@project) && return` works as expected because we have `redirect_to(@project)`. I tested the import functionality manually with a non-CSV file, an empty CSV file, a CSV file that only had an id and title, and a correct CSV file, and for each input, the `redirect_to(@project) && return` works as expected. As an example, here is a screenshot that shows the redirection when I try to import a non-CSV file:
Screen Shot 2021-05-27 at 11 32 07 AM

arielj:
The changes here are pretty safe and I don't think this needs QA, but I think it's better to first merge a PR that fixes the mimemagic issue (it can be fiona's or one specific PR for that) so we get the tests all green before merging?

@fastruby/black-bunny-brigade what do you think?

bronzdoc:
@torresga I think if you rebase main this pr build will pass

torresga:
@bronzdoc done

Issue #23, opened 2021-02-19
[FEATURE] Action Plan Item's prefix


arielj:
Currently, when copying the action plan from points into the roadmap, we have to add the counter/prefix for each item.
From this:
![image](https://user-images.githubusercontent.com/188464/108537906-ebc66d00-72bc-11eb-900c-f226fb0fc1fd.png)

To this:
![image](https://user-images.githubusercontent.com/188464/108537831-d5b8ac80-72bc-11eb-8338-9e235d225e9c.png)

It would be nice that the site adds the numbering automatically and to have an input field at the top of the action plan to add a prefix. For example: all the items are numbered from 1 and on. If the prefix input has some data, it should be prepended to all title (that could be done on the fly with javascript). In the sample screenshots, we could have all the items starting with `1` and so on, and the prefix input with the value `3.1.` to generate `3.1.1`, `3.1.2`, etc..

That way, copy/pasting the action plan would be much easier.

PR #53, opened 2021-06-22
Feature/add action plan prefix

torresga:
**Description:**

This PR closes _#23_ . It adds an input that adds a prefix to each item in the action plan.

Before screenshot: 
Screen Shot 2021-06-22 at 4 19 55 PM

After video:

https://user-images.githubusercontent.com/6892410/122994011-f92c5a80-d375-11eb-84ee-d07d6049f10b.mov

I tested this manually in Chrome, Firefox, and Safari.

### DIFF
--- a/app/assets/javascripts/action_plan.js
+++ b/app/assets/javascripts/action_plan.js
@@ -2,9 +2,8 @@ document.addEventListener("turbolinks:load", function() {
   $("input[name='action-plan_prefix']").keyup(function(event) {
     const prefix = event.target.value.trim();
 
-    $("h3.action-plan_heading").each(function(index) {
-      $(this).children("span").remove();
-      $(this).prepend(`<span>${prefix}${prefix == "" ? "" : index + 1} </span>`);
+    $("h3.action-plan_heading > span").each(function(index) {
+      $(this).text(`${prefix}${prefix == "" ? "" : index + 1 } `);
     });
   });
 });
--- a/app/assets/javascripts/action_plan.js
+++ b/app/assets/javascripts/action_plan.js
@@ -4,7 +4,7 @@ document.addEventListener("turbolinks:load", function() {
 
     $("h3.action-plan_heading > span").each(function(index) {
       const suffix = prefix.length === 0 ? "" : `${index+1}`
-      $(this).text(`${prefix}${suffix} `)
+      $(this).text(`${prefix}.${suffix} `)
     })
   })
 })
--- a/app/assets/javascripts/action_plan.js
+++ b/app/assets/javascripts/action_plan.js
@@ -1,9 +1,10 @@
 document.addEventListener("turbolinks:load", function() {
   $("#action-plan-prefix").keyup(function(event) {
-    const prefix = event.target.value.trim();
+    const prefix = event.target.value.trim()
 
     $("h3.action-plan_heading > span").each(function(index) {
-      $(this).text(`${prefix}${prefix === "" ? "" : index + 1 } `);
-    });
-  });
-});
+      const suffix = prefix.length === 0 ? "" : `${index+1}`
+      $(this).text(`${prefix}${suffix} `)
+    })
+  })
+})
--- a/app/assets/javascripts/action_plan.js
+++ b/app/assets/javascripts/action_plan.js
@@ -6,7 +6,7 @@ document.addEventListener("turbolinks:load", function() {
       let prevHeader = $(this).text();
       prevHeader = prevHeader.substring(prevHeader.indexOf(" ") + 1);
 
-      $(this).text(`${prefix}.${index + 1} ${prevHeader}`);
-    })
+      $(this).text(`${prefix}${prefix.length > 1 ? index + 1 : ""} ${prevHeader}`);
+    });
   });
 });
--- a/app/assets/javascripts/action_plan.js
+++ b/app/assets/javascripts/action_plan.js
@@ -1,10 +1,10 @@
 document.addEventListener("turbolinks:load", function() {
   $("input[name='action-plan_prefix']").keyup(function(event) {
-    let prefix = event.target.value;
+    const prefix = event.target.value.trim();
 
     $("h3.action-plan_heading").each(function(index) {
       $(this).children("span").remove();
-      $(this).prepend(`<span>${prefix}${prefix.trim() == "" ? "" : index + 1} </span>`);
+      $(this).prepend(`<span>${prefix}${prefix == "" ? "" : index + 1} </span>`);
     });
   });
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/action_plan.js
```diff
@@ -0,0 +1,10 @@
+document.addEventListener("turbolinks:load", function() {
+  $("input[name='action-plan_prefix']").keyup(function(event) {
+    let prefix = event.target.value;
```
arielj, line 3: this can be `const prefix` since it's not reasigned

app/assets/javascripts/action_plan.js
```diff
@@ -0,0 +1,10 @@
...
+
+    $("h3.action-plan_heading").each(function(index) {
+      $(this).children("span").remove();
+      $(this).prepend(`<span>${prefix}${prefix.trim() == "" ? "" : index + 1} </span>`);
```
arielj, line 7: you can move the `prefix.trim()` call outside the loop so you don't need to trim the prefix for each iteration

app/assets/javascripts/action_plan.js
```diff
@@ -0,0 +1,10 @@
+document.addEventListener("turbolinks:load", function() {
+  $("input[name='action-plan_prefix']").keyup(function(event) {
+    let prefix = event.target.value;
+
+    $("h3.action-plan_heading").each(function(index) {
+      $(this).children("span").remove();
```
arielj, line 6: what do you think about always having the `span` element there in the html view and set the innerText value instead of removing/prepending it? it should be more performant

you can do something like

```
$('h3.action-plan_heading > span').each(function(span) {
  span.innerText = `${prefix}${prefix.trim() == "" ? "" : index + 1}`;
})
```

arielj (commented):
this is great, I really wanted this feature in the last 2 roadmaps xD

just added some refactor suggestions and it would be great if you can add some system test too

app/assets/javascripts/action_plan.js
torresga: @arielj That is okay by me, especially if it is more performant than my version. I will make that change.

arielj:
hey @torresga, I see the comments were marked as resolved but I see no changes in the code, just checking in

torresga:
> hey @torresga, I see the comments were marked as resolved but I see no changes in the code, just checking in

Oops, sorry, I just hadn't pushed my changes yet. I am working on adding some tests now.

buys-fran:
> > hey @torresga, I see the comments were marked as resolved but I see no changes in the code, just checking in
> 
> Oops, sorry, I just hadn't pushed my changes yet. I am working on adding some tests now.

@torresga will you let me know when you would like me to review or QA?

torresga:
Hi @buys-fran , this is ready for review!

app/views/action_plans/show.html.erb
```diff
@@ -3,8 +3,14 @@
 
   <div id="action-plan">
     <h2>Action Plan</h2>
+
+    <label for="action-plan_prefix">
+      Enter Prefix
+      <input name="action-plan_prefix" type="text">
```
buys-fran, line 9: I think the label's `for` should match the input's `id`
```suggestion
      <input id="action-plan_prefix" name="action-plan_prefix" type="text">
```

app/assets/javascripts/action_plan.js
```diff
@@ -0,0 +1,9 @@
+document.addEventListener("turbolinks:load", function() {
+  $("input[name='action-plan_prefix']").keyup(function(event) {
```
buys-fran, line 2: If you give the input an id, then you can use the id over here
```suggestion
  $("#action-plan_prefix").keyup(function(event) {
```

app/assets/javascripts/action_plan.js
```diff
@@ -0,0 +1,9 @@
...
+    const prefix = event.target.value.trim();
+
+    $("h3.action-plan_heading > span").each(function(index) {
+      $(this).text(`${prefix}${prefix == "" ? "" : index + 1 } `);
```
buys-fran, line 6: I noticed that it is using `==` instead of `===` would it need to coerce the prefix to ""?

app/views/action_plans/show.html.erb
buys-fran: I wonder if you would also prefer using `-` instead of `_` for the id?

app/assets/javascripts/action_plan.js
buys-fran: I wonder if you should also add the (dot) when you append the index?
```suggestion
      const suffix = prefix ? "" : `.${index+1}`
      $(this).text(`${prefix}${suffix } `);
```

buys-fran (commented):
Hey, it looks good. I left a few comments, let me know what you think.

app/views/action_plans/show.html.erb
torresga: @buys-fran I agree with that. I will update the name.

app/assets/javascripts/action_plan.js
torresga: > I wonder if you should also add the (dot) when you append the index?

@buys-fran I was wondering this as well, while looking at https://github.com/fastruby/points/issues/23 I assumed that the expected input would be a prefix that looked like `3.2.`. I am ok with either. It's probably a better user experience to add the dot, what do you think?

buys-fran: I agree, it does seem more user friendly :-)

arielj (approved):
looks good!

buys-fran (commented):
I tested it in chrome. I noticed one very small tweak you could make.
1. Add a prefix
2. Remove the prefix (by pressing backspace)
3. Notice the . (dot) is still there.

I think as a user I would expect the no . (dot) after removing the entire prefix.

image

buys-fran:
> I tested it in chrome. I noticed one very small tweak you could make.
> 
> 1. Add a prefix
> 2. Remove the prefix (by pressing backspace)
> 3. Notice the . (dot) is still there.
> 
> I think as a user I would expect the no . (dot) after removing the entire prefix.
> 
> image

If you think this adds unnecessary complexity, I am okay with proceeding without fixing this edge case. You can remove the dot by refreshing the page.

torresga:
@buys-fran No, I think this is a good edge case. I will update and add tests for this.

buys-fran:
I tested locally and on staging.
Works great 👍

You can merge it, or let me know if you would like me to merge it.

torresga:
@buys-fran Yay! Can you merge it for me please? 😸

PR #54, opened 2021-06-29
Add before action filter to handle proxy requests

buys-fran:
**Description:**

To enable authentication on preview apps we need to do 2 things.

1. A way for auth callbacks to be directed to a preview app
2. A way to provide the preview app url (since they are dynamic)

This PR adds 1. We can proceed to add 2 and only test that 2 works once 1 is deployed.

`The new controller filter will accept a proxy_to param and forward callback requests to the provided domain.`
Rspec specs confirm that this works.

### DIFF
--- a/spec/controllers/callbacks_controller_spec.rb
+++ b/spec/controllers/callbacks_controller_spec.rb
@@ -1,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe CallbacksController, type: :controller do
+  before do
+    @request.env["devise.mapping"] = Devise.mappings[:user]
+    @request.env["omniauth.auth"] =
+      OmniAuth.config.mock_auth[:github] =
+        OmniAuth::AuthHash.new({"extra" => OmniAuth::AuthHash.new({
+          "raw_info" => OmniAuth::AuthHash.new({"login" => "mockuser"})
+        })})
+  end
+
+  describe "#github" do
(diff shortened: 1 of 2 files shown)


spec/controllers/callbacks_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe CallbacksController, type: :controller do
+  before do
+    @request.env['devise.mapping'] = Devise.mappings[:user]
+    @request.env['omniauth.auth'] =
+      OmniAuth.config.mock_auth[:github] =
+        OmniAuth::AuthHash
+        .new({ 'extra' => OmniAuth::AuthHash.new({
```
etagwerker, line 9: @buys-fran Not sure if we are using standardrb or rubocop, but is this the code style we want in the app?

etagwerker (commented):
@buys-fran Looks good. Just had one nitpick about code style. Other than that it looks good.

spec/controllers/callbacks_controller_spec.rb
buys-fran: Thanks for nitpicking. I realised my editor setup was not using standardrb to fix the code.
Now that I got it right, the code looks much better.

etagwerker (approved):

PR #55, opened 2021-06-30
Feature/preview apps redirect pt2

buys-fran:
**Description:**

This PR is related to https://github.com/fastruby/points/pull/54 see that PR for context.

This PR adds a query param to the omniauth callback path.
This query param allows us to redirect back to the preview app after github has authenticated the user.

We also provide a token that goes along with the proxy_to query param. The proxy server can use
the token to verify that the proxy_to url is a valid url.

There is an issue in omniauth that prevents us from adding query params to the callback_path, so I
added a fix and PR to address that issue. FInally I had to then use the fixed commit in our Gemfile.

You can read more about the omniauth issue over here:
 - https://github.com/omniauth/omniauth/issues/661
 - https://github.com/omniauth/omniauth/issues/767
 
And you can see the PR that fixes it over here: https://github.com/omniauth/omniauth/pull/1047

I have added some automated tests. 
And I did test the PR manually. However this is quite hard to do without merging this PR first.

If you wish to do manual QA you would need to update config/initializers/devise.rb so it includes.
```
  if ENV["HEROKU_APP_NAME"]
    OmniAuth.config.full_host = "https://staging.points.ombulabs.com"
    proxy_to = "http://localhost:3000"
    token = Token.issue("#{ENV["PROXY_SECRET"]}#{proxy_to}")
    config.omniauth :github, ENV["GITHUB_APP_ID"], ENV["GITHUB_APP_SECRET"], scope: "user:email",
                                                                             callback_path: "/users/auth/github/callback?proxy_to=#{proxy_to}&token=#{token}"
```

Then update your local .env so it has the correct variables that include all the below.
ORGANIZATION_LOGIN=...
GITHUB_APP_ID=...
GITHUB_APP_SECRET=...
PROXY_SECRET=...
HEROKU_APP_NAME=1

Make sure to deploy this branch on staging.

Finally you will be able to authenticate using staging as the proxy server.

I have already added the PROXY_SECRET to production and review apps environment variables.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -256,7 +256,12 @@ Devise.setup do |config|
   # ==> OmniAuth
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
-  config.omniauth :github, ENV["GITHUB_APP_ID"], ENV["GITHUB_APP_SECRET"], scope: "user:email"
+  if ENV["HEROKU_APP_NAME"]
+    OmniAuth.config.full_host = 'https://staging.points.ombulabs.com'
+    config.omniauth :github, ENV['GITHUB_APP_ID'], ENV['GITHUB_APP_SECRET'], scope: 'user:email', callback_path: '/users/auth/github/callback?proxy_to=http://localhost:3000'
+  else
+    config.omniauth :github, ENV["GITHUB_APP_ID"], ENV["GITHUB_APP_SECRET"], scope: "user:email"
+  end
 
   # ==> Warden configuration
   # If you want to use other strategies, that are not supported by Devise, or
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "token"
+
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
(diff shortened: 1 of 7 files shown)


config/initializers/devise.rb
```diff
@@ -256,7 +258,15 @@
   # ==> OmniAuth
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
-  config.omniauth :github, ENV["GITHUB_APP_ID"], ENV["GITHUB_APP_SECRET"], scope: "user:email"
+  if ENV["HEROKU_APP_NAME"]
+    OmniAuth.config.full_host = "https://points.ombulabs.com"
+    proxy_to = "points-preview-pr-#{ENV['HEROKU_PR_NUMBER']}.herokuapp.com"
```
buys-fran, line 263: I did update the preview app url patter so it matches here.

buys-fran: Here you can see the new pattern when I hover on the link:
image

arielj: do we need this specific url pattern? we should be able to use request.host or something like that I think, that way we can reuse this idea for other projects without requiring the change of the url pattern, what do you think? I know we'll need a similar fix for Dash for example

lib/token.rb
```diff
@@ -0,0 +1,5 @@
+class Token
```
arielj, line 1: I think we should document this a bit, it's not clear why we are using this token. Maybe it needs some documentation here or on the README.md file? like a `Heroku Review Apps Login Proxy` explaining how this works

also, I think it would be better to remove the `algorithm` param, I understand the idea of having flexibility there, but the actual code is not using it and it will be confusing for people reading the code, it's not clear why the param exists when it's only used in the tests, it adds more complexity that is not currently needed. I would go with any hash function for now and only add this configuration if really required in the future

arielj (commented):
hey @buys-fran sorry for the delay on reviewing this

I added some comments and questions in a few places

buys-fran:
We will take a different approach with this issue. 
We are going to try and make use of the Github api instead of adding a reverse proxy on the production server.

I creates unnecessary vulnerabilities and it would also mean that the productions code might need to change for reasons outside of the purpose of the production code.

PR #58, opened 2021-07-15
Upgrade to Rails 6.0 and adds Rails 6.1 via Dual Boot

mateusdeap:
Description:

This PR upgrades our dependencies to Rails 5.2 and gets us closer to Rails 6.0

### TODO

- [x] Set up Circle CI to run a dual boot setup
- [x] Get test suite to pass with Rails 6.0
- [x] Get test suite to pass with Rails 6.1


.circleci/config.yml
```diff
@@ -4,10 +4,10 @@
 #
 version: 2
 jobs:
-  build-rails-5-2:
+  build-rails-6-0:
```
etagwerker, line 7: Considering this will change as new versions are released, I think it would be better to rename it like this: 

```suggestion
  build-rails-current
```

.circleci/config.yml
```diff
@@ -66,10 +66,10 @@ jobs:
           path: /tmp/test-results
           destination: test-results
 
-  build-rails-6-0:
+  build-rails-6-1:
```
etagwerker, line 69: This should probably say something like this: 

```suggestion
  build-rails-next:
```

.circleci/config.yml
```diff
@@ -133,5 +133,5 @@ workflows:
   version: 2
   build:
     jobs:
+      - "build-rails-6-1"
       - "build-rails-6-0"
```
etagwerker, line 137: I believe this should be: 

- build-rails-current
- build-rails-next

etagwerker (changes requested):
One small nitpick re: CI configuration.

bronzdoc (approved):

PR #59, opened 2021-07-19
Updates alert to alert-success

### LINKED ISSUE #57: [BUG] Alert background for errors should not be green
**Before we start...:**

- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository (for multi-repository projects)

**Branch/Commit:**

I have noticed this error on all branches of Points.

**Expected behavior:**

When I see an alert that notifies the user of an error, the background should be a different color (such as red) than a success alert. 

**Actual behavior:**

Currently all alerts have a green background, regardless of whether the message indicates a success or an error. This is confusing because I have to carefully read the alert message instead of relying on the background of the alert for context.

This is an error message, but the alert background is green:
Screen Shot 2021-07-09 at 2 41 17 PM

This is a success message:
Screen Shot 2021-07-13 at 1 59 49 PM

It looks like there is an `alert-danger` class that is being overwritten in the main stylesheet:

Screen Shot 2021-07-13 at 2 05 06 PM

**I will abide by the [code of conduct] (CODE_OF_CONDUCT.md)**

torresga:
**Description:**

This PR closes #57.

While using Points, I noticed that alerts are always a light blue-green color, regardless if the alert represents success or danger. I noticed that the `alert` class was overwritten later in the stylesheet than Bootstrap's contextual classes (`alert-success`, `alert-danger`), which means that because of the CSS cascade, Bootstrap's contextual classes do not display as expected. Instead of overwriting the `alert` class, I just changed that declaration to overwrite the styling for `alert-success` so that the other classes (particularly `alert-danger`) work as expected. 😸 

Screen Shot 2021-07-19 at 12 06 01 PM

### DIFF
--- a/app/assets/stylesheets/3-atoms/_forms.scss
+++ b/app/assets/stylesheets/3-atoms/_forms.scss
@@ -54,7 +54,7 @@ input, textarea {
   }
 }
 
-.alert {
+.alert-success {
   color: black;
   background-color: #D9F1F1;
   border-color: #D9F1F1;


bronzdoc (approved):
LGTM

Issue #50, opened 2021-06-09
Add the possibility to run the app via `docker-compose`


mateusdeap:
**Before we start...:**

- [x] I checked the documentation and didn't find this feature
- [x] I checked to make sure that this feature has not already been requested

**Branch/Commit:**

`main`

**Describe the feature:**

The project would have the necessary configuration files to enable any developer to quickly run it using docker

**Problem:**

As a developer, workflow would be made easier by adding a `Dockerfile` and a `docker-compose.yml` to quickly get the app up and running

**I will abide by the [code of conduct](CODE_OF_CONDUCT.md)**

PR #63, opened 2021-07-22
Dockerize points

bronzdoc:
Closes https://github.com/fastruby/points/issues/50

**Description:**

Adds `docker-compose` development flow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1
 FROM ruby:2.6.3
-RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
+RUN echo "deb http://deb.debian.org/debian/ unstable main contrib non-free" >> /etc/apt/sources.list.d/debian.list
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client firefox xvfb
 WORKDIR /app
 COPY Gemfile /app/Gemfile
 COPY Gemfile.lock /app/Gemfile.lock
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+# syntax=docker/dockerfile:1
+FROM ruby:2.6.3
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
+WORKDIR /myapp
+COPY Gemfile /myapp/Gemfile
+COPY Gemfile.lock /myapp/Gemfile.lock
+RUN bundle install
+
+# Add a script to be executed every time the container starts.
+COPY bin/entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
+ENTRYPOINT ["bin/entrypoint.sh"]
+EXPOSE 3000
+
+# Configure the main process to run when running the image
+CMD ["rails", "server", "-b", "0.0.0.0"]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
+# syntax=docker/dockerfile:1
+FROM ruby:2.6.3
```
arielj, line 2: do tests run with this image? I think it does not include browsers so the feature tests will fail

config/database.yml.sample
```diff
@@ -9,8 +9,10 @@ default: &default
   encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   timeout: 5000
-  host: localhost
-  username: <%= `echo $USER` %>
+  username: postgres
+  password: password
+  host: db
```
arielj, line 14: the values here should be env variables and fallback to strings so it works both with and without docker

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: "3.9"
+services:
+  db:
+    image: postgres
```
arielj, line 4: does it make sense to specify the version that we use here to be the same that we use in production? just in case something changes between session, in production we are using `10.17`

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    volumes:
+      - ./tmp/db:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: password
+  web:
```
arielj, line 9: since we are dual booting, we should have a way to run docker with the next rails version too

arielj (changes requested):
hey @bronzdoc I added some comments here, most of the suggestions are documented here https://app.tettra.co/teams/ombulabs/pages/adding-a-docker-container-to-rails-projects

also, we should update the README to include the steps to run the app using docker

arielj:
@fastruby/black-bunny-brigade tagged in a PR by @bronzdoc

image

docker-compose.yml
bronzdoc: we could do something like  `docker-compose run -e BUNDLE_GEMFILE=Gemfile.next web [COMMAND]`

bronzdoc: Ahh, I got what you mean, somehow I missed the guide link...

lubc:
@arielj do you think you can take a final look at this PR? Thanks!

README.md
```diff
@@ -36,6 +36,27 @@ Go to http://localhost:3000
...
+
+Build the points docker image
+```
+$ docker-compose build
+```
+
+Setup database and run migrations
+```
+$ docker-compose run web bundle exec rake db:setup
+$ docker-compose run web bundle exec rake db:migrate
```
arielj, line 52: Not sure if it's better to do `rails db:setup` instead of `bundle exec rake db:setup`. I know the `rails` command already takes care of loading bundler and I think it does more things. Not too sure though, maybe it's worth the try?

Also, I think `db:setup` already migrates the db (I think it runs `db:schema:load` instead to make it faster) so the `db:migrate` is probably not needed.

README.md
```diff
@@ -36,6 +36,27 @@ Go to http://localhost:3000
...
+```
+
+Run the app
+```
+$ docker-compose up
+```
+
```
arielj, line 59: I think we should also have instructions to run the app in the next rails version.

arielj (commented):
hey @lubc @bronzdoc ! I just added some comments for the documentation to use the docker container

bronzdoc:
I addressed your comments @arielj!

PR #69, opened 2021-08-16
Feature/specify ruby version

### LINKED ISSUE #60: Specify the Ruby version that should be used
**Branch/Commit:**

Main branch

**Describe the feature:**

As a user developer, I want to see which Ruby version I should use to run the app.

**Problem:**

Currently we have a mix of versions:
- CI config is using 2.6.3
- the `.ruby-version.sample` files shows 2.4.1
- the Gemfile has no definition of the Ruby version (Heroku uses this)
- our deployment in heroku is using 2.5.3

Since we are already using 2.6.3 on CI, I think it would be a good idea to set that version to use since it's the one that's always tested (though maybe somebody has another suggestion, maybe test the app in many ruby versions on CI?)

mateusdeap:
**Description:**

Closes #60. Modifies `.ruby-version.sample` to specify ruby 2.6.3, like for the CI. Adds support for asdf by adding a `.tool-versions.sample` and modifying `bin/setup` to create a `.tool-versions` if one is not specified


bronzdoc (approved):
LGTM!

Issue #48, opened 2021-06-08
Sign in is not working in Heroku review apps 


etagwerker:
**Before we start...:**

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)

**Branch/Commit:**

`main` 

**Expected behavior:**

I'd like review apps to let me sign in with my Github username (as it would in production) 

**Actual behavior:**

When trying to sign in I end up in the GitHub 404 page.

**Steps to reproduce:**

1. Step 1: Go to any review app
2. Step 2: Try to sign in

**Screenshots and Videos**

It will redirect to a URL that looks like this: https://github.com/login/oauth/authorize?client_id&redirect_uri=https%3A%2F%2Fpoints-fix-csv-import-e-z5ds36.herokuapp.com%2Fusers%2Fauth%2Fgithub%2Fcallback&response_type=code&scope=user%3Aemail&state=656a13668efdf41a985047c6798ae108eec02eaa124b0bc9

<img width="1362" alt="Screen Shot 2021-06-07 at 9 17 52 PM" src="https://user-images.githubusercontent.com/17584/121107866-b9427080-c7d6-11eb-9b97-5166a64513c0.png">

**Logs**

Include relevant log snippets or files here.

**I will abide by the [code of conduct](CODE_OF_CONDUCT.md)**

buys-fran:
Working on this one

buys-fran:
After lots of research and trying a few things. I have come to the conclusion that there are 3 ways to solve this problem.

1. Create a proxy server, so staging requests can Auth with the same URL production apps auth with.
Not ideal since code meant for staging is deployed to production.
2. Dynamically create a Github APP with the review apps URL as a redirect URL. 
This will require the user to visit the page to create a new app from a manifest. Also not ideal since staging code is deployed to production.
3. Add an alternative login method for staging.

Does anyone mind weighing in on these alrternatives?

buys-fran:
Some links about the github api + creating an app with a manifest.
- https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app-from-a-manifest
- https://docs.github.com/en/rest/reference/apps#create-a-github-app-from-a-manifest

buys-fran:
Some twitter wisdom:
https://twitter.com/searls/status/1012343568072105984

fbuys:
[...]

PR #70, opened 2021-08-16
Updates Modal's event loading so that it always fires

### LINKED ISSUE #65: Add estimates does not always display modal to perform the action 😞
**Before we start...:**

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)

**Branch/Commit:**

main branch

**Expected behavior:**

Clicking add estimate should always display a modal.

**Actual behavior:**

Sometimes clicking add estimate does not display a modal. 

**Steps to reproduce:**

It's hard to find it, but I've experimented it in both Safari and Chrome

1. Step 1: Load a new project page
2. Step 2: Click add estimate 

**Screenshots and Videos**

I will share this privately as it has sensitive details in the video I captured. 

**Logs**

Unfortunately there are no details in the JS console... 😢 

torresga:
**Description:**

Closes #65. 

From what I’ve read from the Turbolinks documentation, we should be using the `turbo links:load` event in place of `DOMContentLoaded` so that the JavaScript is executed on every page change. I think the default behavior with Turbolinks is that `DOMContentLoaded` fires only in response to the initial page load, not after any subsequent page changes, so that’s why the modal only pops up intermittently. (For reference: https://github.com/turbolinks/turbolinks#installing-javascript-behavior)

I ended up testing this manually (going from page to page, pressing the ‘Add Estimate’ button to see if it appears.) If anyone can point me towards some ideas towards testing this through a spec, I would be happy to hear it. 

### DIFF
--- a/app/assets/javascripts/modal.js
+++ b/app/assets/javascripts/modal.js
@@ -1,5 +1,5 @@
 let modal = false
-document.addEventListener('DOMContentLoaded', () => {
+document.addEventListener("turbolinks:load", () => {
   modal = document.querySelector(".modal")
 })
 
--- a/spec/features/modal_displays_always_spec.rb
+++ b/spec/features/modal_displays_always_spec.rb
@@ -1,0 +1,26 @@
+require "rails_helper"
+
+RSpec.describe "managing estimates" do
+  let(:user) { FactoryBot.create(:user) }
+  let(:project) { FactoryBot.create(:project) }
+  let!(:story) { FactoryBot.create(:story, project: project) }
+
+  before do
+    login_as(user, scope: :user)
+  end
+
+  it "allows me to click on the modal twice in a row", js: true do
+    visit project_path(id: project.id)
+    click_link "Add Estimate"
+    expect(page).to have_text("New Estimate")
+    expect(page).to have_content(story.description)
+    find(".close").click()
+
+
+    click_link "points app"
+    find(".project-card:nth-of-type(1)").click()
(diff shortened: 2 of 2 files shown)


bronzdoc:
Thanks for this @torresga! I was trying to QA it now but it seems something is broken with review apps, I will give it a look
Screen Shot 2021-08-16 at 4 30 27 PM

arielj (approved):
Looks good to me, I'm deploying this to staging instead of the review apps since that's not working (check https://github.com/fastruby/points/issues/48)

arielj:
https://staging.points.ombulabs.com/projects

arielj:
> I ended up testing this manually (going from page to page, pressing the ‘Add Estimate’ button to see if it appears.) If anyone can point me towards some ideas towards testing this through a spec, I would be happy to hear it.

@torresga if you want to add an automated test for this, you can probably do the same steps you did for the manual testing with a system spec. Something like (pseudocode):

```
visit("/")
click some_project
click add_estimate
assert that the modal is visible
close the modal
click the title (to go back to the dashboard)
click some_project
click add_estimate
assert that the modal is visible
```

Doing those steps locally before and after the fix was consistent for me to test it, so the test should work (you can remove the fix when adding the test just to be sure fix is fixing the test)

cleicar (approved):
This looks great @torresga. Thank you!

PR #101, opened 2021-10-14
Prevent users interacting with other users' estimates

### LINKED ISSUE #100: Users can edit other users' estimates
**Branch/Commit:**

Main branch

**Expected behavior:**

As a user, I should be allowed to only edit my estimates

**Actual behavior:**

Currently, if I modify the url of the edit estimate page, I can edit any estimate from any user for any project.

**Steps to reproduce:**

1. Go to a project where you have any estimate
2. Right click the `Edit estimate` button and open in a new window (so you have the full url of the edit form)
3. Change the ID of the estimate

They way you can target any estimate from anyone and there's also a problem where the lookup of the estimate is not scoped to the current project/story in the url (we are doing `Estimate.find` instead of something like `current_user.estimates.where(story_id: ...).find(params[:id])` in both the edit/update/destroy actions)

arielj:
**Description:**

This PR fixes a security issue where a user could modify certain parameters in the url to view, edit, update, or destroy estimates made by another user.

I also fixed some things that appeared after changing the code (like the estimate validations were not right), I'll add comments on the diff

Closes #100 

### DIFF
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -8,12 +8,11 @@ class EstimatesController < ApplicationController
   end
 
   def edit
-    @estimate = Estimate.find(params[:id])
   end
 
   def create
-    params["estimate"]["user_id"] = current_user.id
-    @estimate = Estimate.new(estimate_params)
+    @estimate = current_user.estimates.build(estimate_params)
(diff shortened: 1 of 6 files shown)


etagwerker:
Just tested this locally and it's working as expected. 💯

PR #121, opened 2021-10-29
Add helper methods and refactor specs

### LINKED ISSUE #119: Feature specs are too tied to css selectors
**Branch/Commit:**

Main branch

**Describe the feature:**

As a developer, I want to have helper methods to abstract the use of css selectors when writing tests. To make them easier to understand and to write.

**Problem:**

We have things like this:

```rb
select "3", from: "estimate[best_case_points]"
select "8", from: "estimate[worst_case_points]"
```

used everywhere, which is too tied to the code and understaing it requires more knowledge of the context.

**Mockups:**

We could have a helper method like:

```rb
def set_estimates(best, worst)
  select best, from: "estimate[best_case_points]"
  select worst, from: "estimate[worst_case_points]"
end
```

and then we use it like:

```rb
set_estimates("1", "2")
```

This way we can write tests in a more behaviour-driven and declarative way. And if, at some point, the name of the input changes for example, we can update that only in one place.

I have more refactor ideas though, I'll add them in a PR.

arielj:
**Description:**

This PR adds a few helper methods that can be used in feature tests to abstract the CSS selectors from the actual test.

Feature tests should describe what a user would do, so having to write CSS selectors is not ideal. With this helpers, the test can be written in a more user focused way.

closes #119 

___

### DIFF
--- a/spec/features/estimates_manage_spec.rb
+++ b/spec/features/estimates_manage_spec.rb
@@ -21,8 +21,7 @@ RSpec.describe "managing estimates" do
     end
 
     it "allows me to add an estimate" do
-      select "3", from: "estimate[best_case_points]"
-      select "8", from: "estimate[worst_case_points]"
+      set_estimates(3, 8)
       click_button "Create"
       expect(Estimate.count).to eq 1
       expect(page).to have_content "Estimate created!"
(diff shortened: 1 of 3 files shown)


lubc (approved):
@arielj Looks good 👍  Thanks!

PR #122, opened 2021-10-30
Increase line height to prevent title overlap

### LINKED ISSUE #116: Titles look wrong if multiline
**Branch/Commit:**

Main branch.

**Expected behavior:**

The titles should not overlap when multiline and should not be too far apart either.

**Actual behavior:**

If the text is long enough to wrap to a new line, the line spacing looks bad. It's probably just a CSS line-height change.

**Steps to reproduce:**

Set a long project name (or a long project and subproject names), shrink the window width to force it to wrap to 2 lines.

**Screenshots and Videos**

image

taanguyen:
**Description:**
Closes #116 

This fix prevents long project titles from overlapping when the browser window width is reduced. Prior to this fix, when the project title text was long enough to wrap to a new line, the text would overlap. 

Note: I considered using em units, but [MDN Docs say unitless line-height is preferred.](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values) 

## Screenshots
Using the same example text from issue #116

### Before
Screen Shot 2021-10-30 at 12 36 09 AM

### After
Here is the updated screenshot of working functionality (multiline title in the header doesn't cover the project title).
Screen Shot 2021-10-31 at 6 03 27 PM
___

### DIFF
--- a/app/assets/stylesheets/2-quarks/_custom_typography.scss
+++ b/app/assets/stylesheets/2-quarks/_custom_typography.scss
@@ -71,7 +71,6 @@ h4 {
 .dashboard-title-report {
   width: 100%;
   position: sticky;
-  top: 53px;
   background: white;
   text-align: left;
   font-size: 35px;


etagwerker (changes requested):
@taanguyen Thanks for taking a stab at this! 

Your solution works in a way that the header looks good, but the header extends in a way that starts blocking the content from the main container... 

I'm uploading a few screenshots to show the problem: 

Screen Shot 2021-10-30 at 8 19 01 PM

Screen Shot 2021-10-30 at 8 18 45 PM

Screen Shot 2021-10-30 at 8 18 30 PM

It would be great if you can tweak your solution so that if the header is taller, it pushes the content below down...

taanguyen:
Hi @etagwerker! thank you for bringing this to my attention! I'm thinking of adjusting the margin-bottom of the header black bar (that contains the project title). Does this sound good? Or would you suggest an alternative?

etagwerker:
@taanguyen I'm not an expert on HTML/CSS, but I think both <header> and <main> will have to have position:relative and have display:block so that if the <header> is very tall, then <main> will start only after the <header> block finishes... 

I hope that helps!

taanguyen:
Hi @etagwerker! The solution you provided works when the window is about this small (and slightly smaller). 
Screen Shot 2021-10-30 at 7 57 08 PM
But if I squeeze the window to be thinner, the text gets covered again. 
Screen Shot 2021-10-30 at 7 58 32 PM
Any additional suggestions? I tried using margins but they didn't help. Having margins made the display look weird when the window was full size :/

taanguyen:
Hi @etagwerker @lubc! I didn't use the modifications of `position:relative` and `display:block`. Instead, I used 
`@media screen and (max-width: $breakpoint_768px) {
   margin-top: 53px;
  }`
to increase the margin when the window shrinks. This solution is the best that I could come up given my limited knowledge of CSS (I've been trying things unsuccessfully for the last few hours ahah). I noticed several other rules in this file used the `@media` rule so at least, it follows an apparent convention! 

Screen Shot 2021-10-30 at 10 19 15 PM

One thing to note: when you shrink the window even further, the best estimate and worst estimate text overlap and the worst estimate text overlaps with the button! Maybe this could be resolved in a separate issue?

Screen Shot 2021-10-30 at 10 22 20 PM

lubc:
@taanguyen One workaround could be to get rid of the `position: fixed` effect of the header. Ernesto suggested using `position: relative` instead but you can achieve the same by removing the `navbar-fixed-top` class in the header (see https://github.com/fastruby/points/blob/main/app/views/layouts/_header.html.erb#L1)

Once you do that, if there's too much blank space after the header, you can adjust it here https://github.com/fastruby/points/blob/main/app/assets/stylesheets/1-utilities/_basic.scss#L11 to a point where it looks good.

Hope this helps!

taanguyen:
Hi @lubc! Sorry for all the emails. I followed your suggestions and the title doesn't get clobbered by the header anymore- no js necessary! thank you!! I forgot that position: fixed removes the element out of normal document flow. 

The only additional issue that came up was when scrolling, the title and the table get interspersed: 
Screen Shot 2021-10-31 at 5 54 24 PM

I know the CSS has position: sticky so I'm not sure if the intent was to keep the title on top at all times. To stop the title text and table text from jumbling, I made an extra change- I removed top: 53px. So when you scroll down, the two elements don't jumble 
Screen Shot 2021-10-31 at 6 00 58 PM. I hope removing the top offset was ok! Let me know if this was not a desirable modification! 
Here is the updated screenshot of working functionality (multiline title in the header doesn't cover the project title).
Screen Shot 2021-10-31 at 6 03 27 PM

etagwerker (approved):
@taanguyen Looks good, thank you! 👍

taanguyen:
thank you @etagwerker and @lubc!! I had hoped to be of more use but my CSS knowledge is clearly lacking - sorry for all the notifications too. Thank you both for all your help!! 😃

etagwerker:
@taanguyen Thanks for submitting your first contribution to this project and being so quick to respond! ❤️

taanguyen:
@etagwerker Thank you for allowing first-timers like me to contribute and thank you for your help!! 👍  ❤️

PR #124, opened 2021-11-02
Add Github Actions for Points

neutrino:


.github/workflows/points.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.0.2"
+          bundler-cache: true
+
+      - name: Build App
+        run: |
+          yarn install
+          cp config/database.yml.github config/database.yml
```
fbuys, line 47: Why don't we use the existing database.sample.yml?

.github/workflows/points.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.0.2"
+          bundler-cache: true
+
+      - name: Build App
+        run: |
+          yarn install
```
fbuys, line 46: Do we need to run bundle install?

.github/workflows/points-next.yml
```diff
@@ -0,0 +1,55 @@
...
+          ruby-version: "3.0.2"
+
+      - name: Build App
+        run: |
+          bundle install
+          yarn install
+          cp config/database.yml.github config/database.yml
+          bundle exec rails db:create
+          bundle exec rails db:schema:load
+          gem install next_rails
```
fbuys, line 50: I see `next_rails` is included for the :development and :test groups, in which environment does CI run? And does it not have next_rails included from the Gemfile?

fbuys (commented):
Looks good.
I made a few comments, let me know what you think ✅

bronzdoc (commented):
LGTM after @fbuys comments are addressed 👍

.github/workflows/points-next.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/checkout@v1
+
+      - name: Install PostgreSQL 11 client
+        run: |
+          sudo apt-get -yqq install libpq-dev
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.0.2"
```
arielj, line 41: we can add `bundler-cache: true` as an option here so it caches the result of `bundle install`

check this I did in another project https://github.com/arielj/vanilla-nested/blob/main/.github/workflows/main.yml

.github/workflows/points-next.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Build App
+        run: |
+          bundle install
+          yarn install
```
arielj, line 46: we can use a github action that will cache the result of yarn install, check the one I used here https://github.com/arielj/vanilla-nested/blob/main/.github/workflows/main.yml

.github/workflows/points-next.yml
```diff
@@ -0,0 +1,55 @@
...
+          cp config/database.yml.github config/database.yml
+          bundle exec rails db:create
+          bundle exec rails db:schema:load
```
arielj, line 49: we can do `bundle exec rails db:setup` to replace these 2 commands

.github/workflows/points-next.yml
```diff
@@ -0,0 +1,55 @@
...
+          gem install next_rails
+          next bundle install
```
arielj, line 51: since this project is running agains Rails' main branch, we should update the rails version or it will get stuck always testing against the git commit that's in the current .next.lock, we do that in dash https://github.com/fastruby/dash/blob/main/.circleci/config.yml#L114

.github/workflows/points.yml
```diff
@@ -0,0 +1,53 @@
...
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  build-rails-current:
+    runs-on: ubuntu-latest
+    
+    env:
+      POSTGRES_USER: postgres
```
arielj, line 15: I think that we can reuse the configuration from the previous job

if we do `env: &ENV` when we define the first one then we can do

```
env:
  << *ENV
```

here

the sam can be done for the postgres config I think (I'm not 100% sure if that works between jobs, but it's worth to try)

arielj:
Should we remove the circleci config in this PR too? since this is replacing CircleCI

lubc:
hey @neutrino , did you have a chance to look at the comments? Thanks!

arielj (commented):
hey @neutrino I added some more comments, though I think at some points we'll have to merge this to properly test it, right now it's not running the actions so it's hard to verify that it works

also, this PR should remove the circleci config file

.github/workflows/points-next.yml
neutrino: @arielj : That step requires `package-lock.json` to be available in the repository. We don't have `yarn.lock` either. I wander, if we really need `yarn install` here.

.github/workflows/points.yml
neutrino: It's done in the `Setup Ruby` step with `ruby/setup-ruby@v1` but it wasn't working for `next bundle`

config/database.yml.sample
```diff
@@ -9,9 +9,9 @@ default: &default
   encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   timeout: 5000
-  username: <%= ENV.fetch("POSTGRES_USER") { `echo $USER`} %>
-  password: <%= ENV.fetch("POSTGRES_PASSWORD") { "" } %>
-  host: <%= ENV.fetch("POSTGRES_HOST") { "" } %>
+  username: <%= ENV.fetch("POSTGRES_USER") { "postgres" } %>
+  password: <%= ENV.fetch("POSTGRES_PASSWORD") { "postgres" } %>
+  host: <%= ENV.fetch("POSTGRES_HOST") { "localhost" } %>
```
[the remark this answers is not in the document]
neutrino: @arielj : We don't have POSTGRES_USER and POSTGRES_PASSWORD in the `.env.sample` yet.

.github/workflows/points-next.yml
arielj: @neutrino I think we can ignore this in this repo, I'm checking the original CircleCI config and it's not running `yarn install` and the package.json is just empty so running `yarn install` is useless, let's remove that

config/database.yml.sample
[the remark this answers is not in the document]
arielj: @neutrino but those variables are being set by the docker-compose.yml file, though the docker-compose.yml file should also set the POSTGRES_HOST variable

Docker is setting variables so I don't think we need to change the fallback since we can set the variables

arielj:
hey @neutrino can you update the description of the PR describing what it's doing? we shouldn't have blank descriptions

kindoflew:
hi @neutrino! just wondering what the status of this PR is (trying to keep Cycle Time down 💪🏻 ). Thanks!

arielj (approved):
I think this looks good now, we have to merge it to test the whole setup since it's not running yet

Gemfile.next.lock
```diff
@@ -123,8 +124,9 @@ GEM
       thor (~> 1.0)
     builder (3.2.4)
     byebug (11.1.3)
-    capybara (3.35.3)
+    capybara (3.36.0)
```
arielj, line 127: @neutrino are all these updates needed? I see updates for capybara, selenium, omniauth-oauth2, rubocop, seem unrelated to objective of this PR

neutrino:
Ran into the following issue related to the bundle install with the next version of rails
https://github.com/rubygems/rubygems/issues/5088

I tried upgrading the gems and downgrading the bundler version and it's still failing. I will update it once we have a new version.

arielj:
This is replaced with https://github.com/fastruby/points/pull/145

danger89:
> 5088

I also think that #5088 is still not solved. Also using the latest bundler version it fails: https://gitlab.melroy.org/melroy/melroy-site/-/jobs/9158

PR #135, opened 2021-11-25
Initialize position on creation and fix order issue for duplicated positions

### LINKED ISSUE #9: Ordering doesn't work consistently
Sometimes, after dragging a story by mistake (but leaving it in the same spot where it was) or editing stories moves them to different positions.

I couldn't find a concrete step-by-step way to reproduce it.

I think the issue is that the `sort` action increases positions by 1 and new stories my have a lower position

arielj:
**Description:**

This PR fixes #9 by doing three things:
- now, when a story is created, the `position` attribute is initialized as the last position + 1
- since not all stories in the database have a position, this PR adds a secondary order by created_at to prevent edited stories to show up last by being explicit on how to order stories with duplicated position (in the case of the issue, both stories have a nil position)
- for current projects with stories already with `position == nil`, adding a new story would add a position for it and postgres would move the previous stories to the end when ordering by position, so this also adds a `NULLS FIRST` modifier to the order so old stories with nil position are at the top

**How to test/QA:**

Issue #9 describes the step by step method to reproduce the issue.

___

### DIFF
--- a/app/controllers/action_plans_controller.rb
+++ b/app/controllers/action_plans_controller.rb
@@ -3,7 +3,10 @@ class ActionPlansController < ApplicationController
 
   def show
     @project = Project.find(params[:project_id])
-    @project_stories = @project.stories.order(:position, :created_at)
-    @children = Project.where(parent_id: @project.id).includes(:stories).references(:stories).order("stories.position")
+    @project_stories = @project.stories.by_position
+    @children =
+      Project.where(parent_id: @project.id)
+        .includes(:stories).references(:stories)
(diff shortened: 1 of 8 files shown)


lubc (approved):
Looks good 👍

lubc:
@cleicar could you please QA this? Thank you!

arielj:
Specs are failing because of this flaky spec https://github.com/fastruby/points/issues/137

cleicar (approved):
This looks good @arielj !

PR #139, opened 2021-11-25
Fix `delete story` flaky spec

### LINKED ISSUE #137: Flaky `delete story` spec
**Branch/Commit:**

Main branch

**Expected behavior:**

Specs shouldn't be flaky.

**Actual behavior:**

This feature spec in stories_manage_spec.rb:

```
  it "allows me to delete a story", js: true do
    visit project_path(id: project.id)
    accept_confirm do
      click_link "Delete"
    end
    expect(Story.count).to eq 0
  end
```

sometimes fails if the expectation runs too soon before the delete action finishes processing.

**Solution:**

Adding an extra expectation before checking the Stories count should prevent this flakines, something like `expect(page).to have_text "Story was successfully destroyed"` should make capybara wait for the successful deletion first

**Steps to reproduce:**

This is flaky so it's not 100% reproducible, but running that spec may sometimes lead to a failure saying the `Story.count` is expected to be 0 but it's 1.

arielj:
**Description:**

This PR fixes the flaky `delete story` spec by adding some expectation to make capybara wait for the ajax request to finish before checking the stories' count.

**How to QA/test:**

You can run `rspec ./spec/features/stories_manage_spec.rb:38` multiple times, without this fix it will fail some of the times.

Closes #137 

___

### DIFF
--- a/spec/features/stories_manage_spec.rb
+++ b/spec/features/stories_manage_spec.rb
@@ -37,9 +37,14 @@ RSpec.describe "managing stories", js: true do
 
   it "allows me to delete a story", js: true do
     visit project_path(id: project.id)
+
+    expect(page).to have_text story.title
+
     accept_confirm do
       click_link "Delete"
     end
+
+    expect(page).not_to have_text story.title
     expect(Story.count).to eq 0
   end
 


lubc (approved):
Looks good 👍 Thanks!

PR #162, opened 2021-12-06
Collapse row actions in a 'more' button and restyle

### LINKED ISSUE #140: Collapse stories actions in the rows
**Branch/Commit:**

Main branch

**Describe the feature:**

The actions for each story in the project's show page should use less space so there's more room for the story title and to make it look less bloated

**Problem:**

This is what the page looks like now (before adding the `Move to` button from another issue, when that gets merges this will be even worse)
image

And the layout starts looking really bad with not that small screens already
image

**Mockups:**

We have 2 options:
- we can replace buttons with icons (trash can to delete, pencil to edit, a copy icon for the clone, etc)
- we can only show the most used buttons (Edit and Add/Edit Estimate?) and group the rest of the actions under a `...` button to show the rest of the options

I personally prefer the idea of having icons, I think it looks cleaner and we don't need an extra click on a `...` icon to show more options (which makes it hard to discover new actions for example)

If we go with icons, we are already uses font-awesome so we could have:
- Clone https://fontawesome.com/v4.7/icon/files-o
- Delete https://fontawesome.com/v4.7/icon/trash-o
- Edit https://fontawesome.com/v4.7/icon/pencil-square-o
- Add/Edit Estimate https://fontawesome.com/v4.7/icon/signal maybe?
- Move to https://fontawesome.com/v4.7/icon/share-square-o

arielj:
**Description:**

This PR is a restyle of the row's actions in the project's show page. The current group of actions is too big with many option options that are not used too often (like clone, destroy, move to). I also changed the white over green buttons to the less blinding black over yellow colors.

After checking with nathalie, we decided the best UX is to only show the button with text for the main action (add/edit estimates) and group the rest of the actions in a dropdown, instead of showing all the options as buttons as described in the linked issue.

There are other suggestions to improve the colors and hovers style of buttons but I'll add some issues to address those suggestions.

image

`more actions` dropdown
image

`move to` dropdown
image

Closes #140 

___

### DIFF
--- a/app/assets/stylesheets/4-molecules/_tables.scss
+++ b/app/assets/stylesheets/4-molecules/_tables.scss
@@ -11,7 +11,7 @@
   }
   .project-table__row {
     display: grid;
-    grid-template-columns: 40% 10% 10% auto;
+    grid-template-columns: 1fr 70px 70px 260px;
     align-items: center;
     padding: 10px 0;
     &.project-table__row--reports {
--- a/app/assets/stylesheets/project.scss
+++ b/app/assets/stylesheets/project.scss
@@ -75,26 +75,57 @@
   }
 }
 
-.move-story-wrapper {
+.dropdown-wrapper {
   display: inline-block;
   position: relative;
 
-  .move-story-dropdown {
-    opacity: 0;
-    transition: opacity 0.3s;
-    pointer-events: none;
+  .dropdown {
+    display: none;
     position: absolute;
     right: 0px;
-    top: 100%;
+    top: 50%;
     z-index: 2;
+    min-width: 150px;
+    background: white;
+    box-shadow: 1px 1px 4px 2px rgba(0, 0, 0, 0.4);
 
-    input[type="submit"]:hover {
-      background: lightgray;
+    a,
+    button,
+    input {
+      width: 100%;
+      border: 0px;
+      margin: 0px;
+      padding: 0.5rem;
+      background: none;
+      text-align: left;
+      font-size: 1.2rem;
+      font-weight: normal;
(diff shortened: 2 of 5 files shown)


kindoflew:
This looks good to me! I was following along in Slack, and I like this dropdown style better as well!

My only thought is to maybe move the nested `move to` dropdown off to the side, instead of on top of the other dropdown? But that might be nitpicking 🤔 .

Hi @mateusdeap! Could you QA this when you have a chance? Thanks!

kindoflew (approved):

arielj:
> My only thought is to maybe move the nested `move to` dropdown off to the side, instead of on top of the other dropdown? But that might be nitpicking thinking .

I changed it :+1:

image

lubc:
@arielj could you please fix the git conflict? thanks!

lubc:
@mateusdeap hey, when you have some time please do the QA of it. Thank you!

mateusdeap:
So I noticed something and I'm not sure if it's a pre existing bug or introduced in this change.

When I create a story, the Add Estimate button is yellow
image

Once I add it, it becomes Edit Estimate and it's green
image

If I open the dropdown and edit the story
image

Once I'm done, it's yellow again.
image

Is this expected?

mateusdeap:
Otherwise, this looks good to me. @arielj Just want some feedback on my previous comment before approving

arielj:
@mateusdeap I pushed a fix for the button's color after adding an estimate, it shouldn't turn green now

mateusdeap (approved):

mateusdeap:
LGTM

cleicar (approved):
Looks great! 🚀

PR #165, opened 2021-12-10
Allow user to sort sub projects

### LINKED ISSUE #111: Add the ability to order sub-projects
**Branch/Commit:**

Main branch

**Describe the feature:**

As a user, I want to be able to re-order the sub-projects of a given project, so when I generate the complete action plan, the projects are in the expected order.

**Problem:**

Currently, the sub-projects of a given project are ordered in the creation order, if a new sub-project needs to be before a previous sub-project when generating the full action plan, that can't be done right now.

**Mockups:**

I imagine to use something like we do now with drag/drop of stories but for the list of subprojects, but I don't have a mock or a concrete idea.

neutrino:
**Description:**
Allows user to sort sub-projects.

*** How to test/QA: ***
Create a project and few sub-projects under it. The sub-projects can me move around to the desired position. 

Note: This PR also changes the routes for sorting stories and adds few specs for sub-project feature. 

Closes #111


arielj (commented):
hey @neutrino I know you were doing some tests with the `drag_to` capybara method, it didn't work for this? I think it would be nice to add a test (you added some tests, but not for the feature added here)

bronzdoc:
@neutrino Can you address Ariel comments? thanks!

neutrino:
@bronzdoc @arielj : I was playing with the `drag_to` method to add the test for the drag/drop feature spec. It wasn't adding the sortable css class and was stuck with that.

lubc:
@arielj Could you take another look at this PR when you have some time? Thanks!

arielj:
merging into `sort-sub-project` to continue the work

PR #166, opened 2021-12-10
Prevent the creation of multiple estimates

### LINKED ISSUE #142: Users can create multiple estimations for the same story
**Branch/Commit:**

Main branch

**Expected behavior:**

If a user already estimated a story, creating an estimate should either fail or edit the current estimation instead of creating new estimations. Also, trying to add an estimate for a story with an estimate already should redirect to the edit estimate form instead.

**Actual behavior:**

You can open the `Add Estimate` form in multiple tabs and create as many estimates as you want for a story. You can also open the Add Estimate with the url and add more estimates.

This is closely related to https://github.com/fastruby/points/issues/141 (that story is about preventing the problem generated by this bug)

**Steps to reproduce:**

- Go to a project with a story that you didn't estimate yet
- Open the Add Estimate action in a new window, multiple times
- Submit estimates for each tab

arielj:
**Description:**

This PR closes #142 by checking if there's currently an estimate for a story for the given user before creating a new estimate. If an estimage for the [story, user] already exists, the create action updates it instead of creating an new one.

___

### DIFF
--- a/spec/controllers/estimates_controller_spec.rb
+++ b/spec/controllers/estimates_controller_spec.rb
@@ -3,10 +3,10 @@ require "rails_helper"
 RSpec.describe EstimatesController, type: :controller do
   render_views
 
-  let!(:user) { FactoryBot.create(:user) }
-  let!(:project) { FactoryBot.create(:project) }
-  let!(:story) { FactoryBot.create(:story, project: project) }
-  let!(:estimate) { FactoryBot.create(:estimate, story: story, user: user) }
+  let(:user) { FactoryBot.create(:user) }
+  let(:project) { FactoryBot.create(:project) }
+  let(:story) { FactoryBot.create(:story, project: project) }
+  let(:estimate) { FactoryBot.create(:estimate, story: story, user: user) }
 
   before do
     @request.env["devise.mapping"] = Devise.mappings[:user]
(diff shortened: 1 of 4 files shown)


spec/controllers/estimates_controller_spec.rb
```diff
@@ -3,10 +3,10 @@
...
   render_views
 
-  let!(:user) { FactoryBot.create(:user) }
-  let!(:project) { FactoryBot.create(:project) }
-  let!(:story) { FactoryBot.create(:story, project: project) }
-  let!(:estimate) { FactoryBot.create(:estimate, story: story, user: user) }
+  let(:user) { FactoryBot.create(:user) }
+  let(:project) { FactoryBot.create(:project) }
+  let(:story) { FactoryBot.create(:story, project: project) }
+  let(:estimate) { FactoryBot.create(:estimate, story: story, user: user) }
```
arielj, line 9: I removed the `!` since some of the tests require the estimate to not exist now that we have the uniquness validation

kindoflew:
hi @cleicar! would you mind reviewing this when you have a chance? thanks!

bronzdoc (approved):

cleicar (changes requested):
@arielj I was QAying this but I was able to create multiple estimates for the same story. This is the project that I was using for the tests: https://points-staging.herokuapp.com/projects/2

Screen Shot 2021-12-18 at 17 53 17

arielj:
> @arielj I was QAying this but I was able to create multiple estimates for the same story. This is the project that I was using for the tests: https://points-staging.herokuapp.com/projects/2
> 
> Screen Shot 2021-12-18 at 17 53 17

@cleicar how did you verify that 2 estimates were created? I'm checking the database and there's only one estimate for your user and that story

this won't prevent a user to open multiple `Add estimate` screens and submit the form multiple times, it takes care of that at the controller level, it updates a previous estimate if one is present instead of creating a new one (the message still says an estimate was created because it was sent to the create action, but it doesn't create more than 1 estimate)

cleicar (approved):
@arielj I think that what confused me is the displayed message, but what you said make sense. Let's leave it like that.

I can confirm that submitting a second form updates the estimates, so this is good to go!

PR #177, opened 2022-01-05
Show the project status with the project name

### LINKED ISSUE #176: Show any indication that the project is archived
Currently there is no clear indication to show if the project is archived or not, other than the "Unarchive Project" button.

For a better user experience would be nice to have a label or any indication showing that the project is archived either in the projects page and in the project details page.

**Branch/Commit:**
`main`

cleicar:
Description:

This closes #176 and shows the project's status with the project name so it's easy to identify archived projects.

Screen Shot 2022-01-05 at 16 35 17

It is also showing the same status bagde in the projects page: 
Screen Shot 2022-01-05 at 16 35 06

___

### DIFF
--- a/app/assets/stylesheets/2-quarks/_custom_typography.scss
+++ b/app/assets/stylesheets/2-quarks/_custom_typography.scss
@@ -76,6 +76,10 @@ h4 {
   font-weight: 300;
   line-height: 40px;
   z-index: 1;
+
+  .status-badge {
+    font-size: 16px;
+  }
 }
 
 .fixed-header {
--- a/app/assets/stylesheets/3-atoms/_badges.scss
+++ b/app/assets/stylesheets/3-atoms/_badges.scss
@@ -1,12 +1,12 @@
 .status-badge {
   display: inline-block;
-  padding: 0.25em 0.4em;
+  padding: 10px 20px;
   font-weight: 700;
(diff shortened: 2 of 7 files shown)


lubc:
@neutrino hey, could you please review this PR? Thank you!

kindoflew:
hey @KattKit, could you review this when you have a minute? thanks!

KattKit (approved):
I'm not going to hold this up, so my two cents and comments on this are:

1) I would expect the badge to look more like the buttons on the page, as it stands its very rectangular and kind of small on the padding: I don't think the text itself needs to be near the size of the title but it looks odd in context and I think expanding the padding can generally catch the eye better as well as making the status badge feel less out of place

2) more generally: are we expecting this `status` field to expand in any way? In other projects archiving was done with a boolean called `archived`, which was really useful because when there were feature request for new badges like this one and related data to track, because we could simply add a boolean column and add the badge to the page as a boolean.

cleicar:
Thanks for your feedback @KattKit! 

I did the style changes in the badge and this is how it looks now:

Screen Shot 2022-01-11 at 18 58 07

About the `status` field, I believe that we plan to keep it but it might better to move the discussion outside of this PR scope.

bronzdoc (approved):

PR #179, opened 2022-01-07
Advanced project cloning

### LINKED ISSUE #148: Customize cloning a project
**Branch/Commit:**

Main branch.

**Describe the feature:**

As a user, I would like to click `Clone Project` and have some options to configure how the clone is created:
- set the name of the new project
- should it be a sub-project? if so show a selector of current projects
- should it also clone sub-projects? (it shouldn't clone the sub-projects if it's cloned as a sub-project, since we don't allow more than 1 nesting level)

**Problem:**

Currently, cloning a project creates a new project with a fixed name `Copy of ...`, it only copies the parent project but not the sub-projects and you can't re-assign the parent project.

Also, if you click `Clone Project` by mistake, it clones the project right away which is not ideal, with this intermediate screen to customize the clone we can prevent that mistake too.

arielj:
**Description:**

This PR adds a screen to configure how a project is cloned. Now we can pick the new title and the parent project.

Also, cloning a project with sub projects will also clone the sub projects and its stories.

image

**How to QA:**

Try cloning different projects (parent projects and sub projects) selecting different configurations for the parent and title.

Closes #148 
___


app/controllers/projects_controller.rb
```diff
@@ -26,16 +26,18 @@ def toggle_archive
...
-    original = Project.includes(stories: :estimates).find(params[:id])
-    duplicate = original.dup
-    duplicate.title = "Copy of #{original.title}"
-    duplicate.save
+  def clone
+    @original = Project.includes(:projects, stories: :estimates).find(params[:id])
+  end
 
-    original.stories.each { |x| duplicate.stories.create(x.dup.attributes) }
+  def do_clone
```
cleicar, line 33: @arielj one suggestions for this methods named: `new_clone` and `clone`.

cleicar (approved):
@arielj great refactor replacing `duplicate` with `clone`!

I have just a small comment, but otherwise it's approved.

app/controllers/projects_controller.rb
arielj: thanks! I updated the names of the actions

bronzdoc (approved):

bronzdoc:
@mateusdeap can you QA this?

lubc:
@mateusdeap 👀

kaysiz:
@lubc I can QA this, I have some time

kaysiz (approved):
QA LGTM. I was able to clone a project with a sub-project with no parent project and also with a parent project which could be itself. However, I am not sure if it is by design that we do not validate the project name to be unique.  🚀

PR #185, opened 2022-01-13
Run feature tests using JavaScript

torresga:
**Description:**

This PR runs all the feature tests with JS by default. I initially deleted the tests that didn't pass with `js: true` but then I added them back per @arielj 's suggestion. We also refactored the "allows me to update existing stories on import" test by removing the helper methods `download_csv_file` and `adjust_csv_descriptions` because we thought those methods were too complicated and it didn't make sense to test downloading the CSV file in that test.  _Closes #118_

### DIFF
--- a/spec/features/stories_manage_spec.rb
+++ b/spec/features/stories_manage_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe "managing stories", js: true do
     expect(page).to have_content "Story updated!"
   end
 
-  it "allows me to delete a story", js: true do
+  it "allows me to delete a story" do
     visit project_path(id: project.id)
 
     expect(page).to have_text story.title
--- a/spec/features/estimates_manage_spec.rb
+++ b/spec/features/estimates_manage_spec.rb
@@ -28,6 +28,13 @@ RSpec.describe "managing estimates", js: true do
       expect(page).to have_content "3"
       expect(page).to have_content "8"
     end
+
+    it "allows me to add an estimate", js: false do
+      set_estimates(3, 8)
+      click_button "Create"
+      expect(Estimate.count).to eq 1
+      expect(page).to have_content "Estimate created!"
+    end
   end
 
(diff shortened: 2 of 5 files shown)


spec/features/stories_manage_spec.rb
```diff
@@ -39,7 +39,7 @@
     expect(page).to have_content "Story updated!"
   end
 
-  it "allows me to delete a story", js: true do
+  it "allows me to delete a story" do
```
bronzdoc, line 42: We don't want to test these with Javascript disabled as in `estimates_manage_spec.rb` ?

torresga: I tried to write a test to delete a story without JavaScript enabled, but it doesn't look like deleting stories works without JavaScript - when I try to delete a story in my browser with JS disabled, I get the following error: 
Screen Shot 2022-01-18 at 1 39 53 PM

kindoflew:
@torresga looks like there is a merge conflict, i'll ping someone to review when it's resolved! thanks!

kindoflew:
hi @mateusdeap! could you review this when you have a minute? thanks!

kindoflew:
@mateusdeap :shipit:

mateusdeap:
Aaaaah Sorry.

mateusdeap (commented):
Seems good to me. Unless we want to clean up the git history for this PR. @kindoflew @torresga I can pair on that. But if that's a minor concern, lgtm

torresga:
> Seems good to me. Unless we want to clean up the git history for this PR. @kindoflew @torresga I can pair on that. But if that's a minor concern, lgtm
Yeah I would appreciate someone taking a look with me to clean up the git history/fix these merge conflicts. I thought I was fixing them but it seems like I am making a bigger mess.

mateusdeap:
Just a flaky spec

mateusdeap (approved):

PR #190, opened 2022-01-16
Upgrade to Ruby 3.1 and Rails 7.0 

etagwerker:
Hey,

This PR upgrades the project to Ruby 3.1 and Rails 7.0

Please check it out.

Thanks! 

___

I will abide by the [code of conduct](CODE_OF_CONDUCT.md).

this is the internal jira link https://ombulabs.atlassian.net/browse/DT-222


lubc (commented):
Looks like the build is failing in `rails-next`
I see there're some deprecation warnings in  the `rails-current` build. Maybe that's related

rishijain:
Hey, @etagwerker thank you for the work on this PR. 
Upon reviewing this PR, I realized maybe it is not the best idea to upgrade Rails and Ruby in the same PR. There are some spec failures as well in this pr which seems to be due to the ruby upgrade.

So to keep things simple to implement, review and test, I am closing this PR and have raised another PR for just the rails upgrade to 7.0 https://github.com/fastruby/points/pull/248

and I have also raised an issue for the ruby upgrade so we can take that up as well https://github.com/fastruby/points/issues/249

The Rails upgrade PR came up nicely due to your work acting as a very good reference point for me to do the upgrade and I am sure it will help the same while upgrading ruby to 3.1. So thank you for your work on this. :heart: 

Let me know if you have any questions or concerns about the PRs or the approach taken for this, would be happy to answer.

Issue #192, opened 2022-01-17
Long links break the new-story form design


arielj:
**Branch/Commit:**

Main branch

**Expected behavior:**

The textarea with the description should have a minimum width so it's always readable.

**Actual behavior:**

If the description includes a super long link, it takes a lot of space, compressing the space for the textarea.

Some `min-width` CSS should fix this

**Steps to reproduce:**

Write a super long link in a description.

**Screenshots and Videos**

![image](https://user-images.githubusercontent.com/188464/149810345-5f43eda5-80c7-46b1-962d-a9afad4b7d9b.png)

**I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)**

PR #194, opened 2022-01-20
Prevent overflowing links in the story preview

fbuys:
**Description:**

Fixes: https://github.com/fastruby/points/issues/192

Long links stretch the with of the preview when they overflow. The input
side of the edit screen shrinks to accommodate the overflow. This is not
ideal because it becomes harder and harder to see your input as it is
reduced in size.

By using minmax function we can be sure that the input size will be at
least 50% of the parent width.

Setting a min width on the input side is not all that is required
because it still allows the long link to overflow requiring that the
user scroll from left to right on the screen to view the entire preview.

So to improve on the minmax solution we also force long links to break
when ever they reach the container boundary.

## For QA
1. Create a story
2. Make sure to include a long link such as: 
http://click.revue.email/ss/c/TzfyQfvFfgo-vTkXDZQg11ss1UDf9kKcltraveMoLkgId13tgDEYAkZth_9V0UDjXlJYGxwIQskCeZmw7nj6QNAyFM6kgMneTr-fcQT0UviQErJjWysr4ikyIabEtAc_aQJZrx9S21p8NtCE91S4QYjxBXUKPBycsIDGGoKxfnI1rv72mf6SRQgnAVEueBX3C9oXwoWPkE619D__G0rOGWXaJs8cDFKdEd-IwOaVvA8/3jk/tAw5AKjuQM-Q3_4kWlAdcw/h28/dO58uY0wnKGR-Zj3sqBfVq9SoTBjZQruGJP-NVaMgSQ
3. Make sure the edit screen still uses 50% of the page for editing and 50% of the page for previewing.

___

### DIFF
--- a/app/assets/stylesheets/stories.scss
+++ b/app/assets/stylesheets/stories.scss
@@ -52,7 +52,7 @@
     "title preview"
     "description preview"
     "submit .";
-  grid-template-columns: 1fr 1fr;
+  grid-template-columns: repeat(2, minmax(50%, 1fr));
   grid-column-gap: 10px;
 
   #error_explanation {
@@ -76,6 +76,10 @@
     .content {
       overflow: auto;
       max-height: min(50vh, 700px);
+      // prevent long links from overflowing
+      a {
+        word-break: break-all;
+      }
     }
   }
 


cleicar (commented):
@fbuys It looks like this change fixes the issue with the long links but I believe we still need to define a min-width to the text are to avoid it to be narrow by any other situation like this one.

fbuys:
> @fbuys It looks like this change fixes the issue with the long links but I believe we still need to define a min-width to the text are to avoid it to be narrow by any other situation like this one.

Thanks @cleicar  that is a very good point. I have updated the solution. Could you take a look again please?

cleicar (approved):
Looks great @fbuys, thanks!

kindoflew:
hi @kaysiz! could you QA this when you have a chance? thanks!

kindoflew:
@fbuys for what its worth, I opened a PR a few days ago that does a very similar thing to the `grid-template-columns` rule. i did it this way:
```
grid-template-columns: repeat(2, minmax(0, 1fr));
```

I learned it here: https://css-tricks.com/equal-width-columns-in-css-grid-are-kinda-weird/.

I'm not sure which one is nicer, or if it's just a matter of opinion. ¯\\_(ツ)_/¯

kindoflew:
@kaysiz, just a reminder to QA this when you can! thanks!

kaysiz (approved):
QA Looks good
Screenshot 2022-02-08 at 13 09 19

fbuys:
> @fbuys for what its worth, I opened a PR a few days ago that does a very similar thing to the `grid-template-columns` rule. i did it this way:
> 
> ```
> grid-template-columns: repeat(2, minmax(0, 1fr));
> ```
> 
> I learned it here: https://css-tricks.com/equal-width-columns-in-css-grid-are-kinda-weird/.
> 
> I'm not sure which one is nicer, or if it's just a matter of opinion. ¯\_(ツ)_/¯

I am a bit late. But it seems like making the lower bound 50% explicitly is clearer? Perhaps using 50% is also possible because we know it is only two columns. Using minmax(0, 1fr) is definitely the answer when you don't know how many columns you might have.

I am happy to use  `minmax(0, 1fr)` but to me it seems like you need to have some understanding of the grid to know that it won't shrink all the way down to 0px.

kindoflew:
@fbuys, good points!! 

in my head, a thing that could also cause confusion is that we want the columns to be equal widths, but `minmax(50%, 1fr)` is only set for one of them.

i guess we could split the difference:
```
grid-template-columns: repeat(2, minmax(50%, 1fr));
```

but at this point i might be bikeshedding 😅. i defer to your wisdom!

fbuys:
@kindoflew I added the changes you suggested. I think it looks good thanks 👍
We should be ready for the next step (not sure if we go back to review/QA)

kindoflew:
@fbuys I don't think it needs another QA, because I think it's just a syntax change -- both rules do the same thing. But I can ping someone, if you prefer!

fbuys:
Lets mergit! :-)

bronzdoc:
@fbuys @kindoflew if you think something it's ready to merge, Just merge it :shipit:

PR #205, opened 2022-02-01
Fix flaky sort sub project spec (WIP)

### LINKED ISSUE #204: Flaky sort sub projects spec
**Branch/Commit:**

Main branch

**Expected behavior:**

Specs should not be flaky.

**Actual behavior:**

The spec to sort sub project uses the drag_to method that is not too stable, causing failures too often on CI.

**Steps to reproduce:**

Check CI for the latest PRs

arielj:
**Description:**

Seems like the drag_to capybara method is not that consistent/stable and it doesn't drop elements in the same position (it gets the center of both elements and moves things with some delay and the sortable javascript needs things to be move a bit more)

Instead of checking the exact order, changed the test to verify the call to the `sort` action was done and then I added a controller test for the sort action.

I re-ran the tests like 5 times and it's been green on all runs

Closes #204 

___

### DIFF
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -210,4 +210,17 @@ RSpec.describe ProjectsController, type: :controller do
       end
     end
   end
+
+  describe "PATCH sort" do
+    let!(:sub_project1) { FactoryBot.create(:project, parent: project, position: 1) }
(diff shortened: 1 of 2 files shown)


kindoflew (approved):

kindoflew:
LGTM! Good call on splitting it into two different tests!

PR #213, opened 2022-02-21
Project Title inline edit

### LINKED ISSUE #149: Edit project title inline
**Branch/Commit:**

Main branch

**Describe the feature:**

As a user, I would like to edit the title of the project from the show page itself instead of clicking `Edit or Delete project`. Somewhat like github does on PRs and Issues where you can edit the title from the same screen.

**Problem:**

Currently, the `edit` action only has the title to be edited and it's also not the best place to have that `delete` button.

**Mockup:**

I imagine something similar to github, an `Edit` button on the right corner in the same like as the title. Maybe we need to move the `parent` project up so there's more place for the text input in case the parent has a long name?

arielj:
**Description:**

Currently, the only editable field of a project is the Title, to edit that we have to go to the Edit or Delete Project action, it requires a lot of clicks just for that and the Edit or Delete Project action has to deal with 2 tasks.

This PR fixes that by adding the option to edit the title field inline in the Show action of the projects. After editing the title it updates the navbar and sidebar when needed too.

With that change I was able to change the Edit or Delete Project button to be only a Delete Project button.

I also changed the layout a bit to fit all the content, if a project has a parent, the parent's title now appears smaller above the current project's title.

This inline-edit is only available in the show action.

https://user-images.githubusercontent.com/188464/154997698-48e3ae15-8a36-477b-9429-6f69ea1fe132.mp4

**How to QA:**

Go to the show action of different projects and click the title to edit it, try different options with that. Also check that it's not editable in other places using the same title.

Closes #149

___


arielj:
I don't know how to fix the code climate thing, I followed a pretty standard rails pattern with the different format.X blocks :man_shrugging:

lubc:
@torresga Can you review this one when you have some time? Thanks!

torresga (approved):

torresga:
This looks good to me, but is there anything we can do to fix the codeclimate build?

arielj:
> This looks good to me, but is there anything we can do to fix the codeclimate build?
I think we can tell codeclimate to ignore it :squirrel:

bronzdoc (approved):

PR #227, opened 2022-03-04
Issue #225 - fixes order of sub projects on action plan page

### LINKED ISSUE #225: Action plan doesn't respect the sub-projects' order
**Branch/Commit:**

Main branch

**Expected behavior:**

When generating an action plan from the parent project, each sub-project should be in the order they are specified.

**Actual behavior:**

The order of sub-projects is not respected

**Steps to reproduce:**

I don't have specific steps, but the order seems to not be correct in the code, it's based on the stories' positions and not on the sub-projects' positions.

mcelaney:
Closes #225

**Description:**

Sub projects on the action plan page are ordered by the position values of their stories. This throws the order of projects off - so we needed to add the project position to the order call.

** What I did **

- Move the project query from the controller to a scope on the Project model
- Write a test for the scope that fails
- Fix the scope.

___

jira link https://ombulabs.atlassian.net/browse/DT-219

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -12,7 +12,7 @@ class Project < ApplicationRecord
 
   scope :active, -> { where.not(status: "archived").or(where(status: nil)) }
   scope :parents, -> { where(parent: nil) }
-  scope :with_ordered_stories, ->(project_id) {
+  scope :sub_projects_with_ordered_stories, ->(project_id) {
     where(parent_id: project_id)
       .includes(:stories).references(:stories)
       .order("projects.position ASC, stories.position ASC NULLS FIRST")
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -12,6 +12,11 @@ class Project < ApplicationRecord
 
   scope :active, -> { where.not(status: "archived").or(where(status: nil)) }
   scope :parents, -> { where(parent: nil) }
+  scope :with_ordered_stories, ->(project_id) {
+    where(parent_id: project_id)
+      .includes(:stories).references(:stories)
+      .order("projects.position ASC, stories.position ASC NULLS FIRST")
+  }
 
   def best_estimate_total
     stories.includes(:estimates).sum(&:best_estimate_average)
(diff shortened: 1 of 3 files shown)


bronzdoc (approved):
LGTM just a small comment

bronzdoc:
@KattKit can QA this? thanks!

app/models/project.rb
```diff
@@ -12,6 +12,11 @@ class Project < ApplicationRecord
 
   scope :active, -> { where.not(status: "archived").or(where(status: nil)) }
   scope :parents, -> { where(parent: nil) }
+  scope :with_ordered_stories, ->(project_id) {
```
arielj, line 15: maybe the name of the scope is a bit confusing? the name is `with_ordered_stories` but it's ordering the projects and I have the feeling that it's not actually ordering the stories inside each project

spec/models/project_spec.rb
```diff
@@ -21,6 +21,19 @@
     expect(sub_project2.position).to eq 2
   end
 
+  describe ".with_ordered_stories" do
+    it "orders sub projects properly" do
+      parent = FactoryBot.create(:project)
+      sub_project1 = FactoryBot.create(:project, parent: parent)
+      FactoryBot.create(:story, project: sub_project1, position: 2)
+      sub_project2 = FactoryBot.create(:project, parent: parent)
+      FactoryBot.create(:story, project: sub_project2, position: 1)
```
arielj, line 30: I think the test is not correct, the problem we have is that we can have multiple stories between all the subprojects with the same position (like story 1 in subproject 1 and story 1 in subproject 2), it's working in this case because we are setting 2 different positions but in 2 real projects we would have a lot or repeated positions (though unique in the scope of each subproject)

This should order subprojects by the sub-project's position field, not the stories one

arielj (changes requested):
hey @mcelaney I think the test is not testing the correct issue, check my comment, also I think we should remove the order on stories for the scope, I don't think it will order the stories as it sounds it would do (maybe we need more tests for that, I'm not sure how to set the order of eager loaded associations)

mcelaney:
@arielj Changed the name for clarity (good point there) and added more examples to the tests to prove the stories are ordering within the sub projects.

kindoflew:
hi, @arielj! have you had a chance to look at the most recent changes to the tests? thanks!

bronzdoc (approved):

PR #231, opened 2022-03-05
Close 152 by preventing sub projects from being archived

### LINKED ISSUE #152: Sub-projects shouldn't allow being archived individually
**Branch/Commit:**

Main branch

**Expected behavior:**

As a user, I shouldn't see the option to archive a sub-project since we don't really handle archived sub-projects in a different way.

**Actual behavior:**

There's a button to `archive` a sub-project, but it's always shown in its parent and is useless.

mcelaney:
Closes #152

**Description:**

Prevent a sub project from being archived separately from their parent project.

___

jira link https://ombulabs.atlassian.net/browse/DT-217

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -46,6 +46,30 @@ RSpec.describe Project, type: :model do
     end
   end
 
+  describe "#archived?" do
+    it "returns true if archived" do
+      expect(Project.new(status: "archived").archived?).to eql(true)
+    end
+
+    it "returns false if status is not `archived`" do
+      expect(Project.new(status: nil).archived?).to eql(false)
+    end
+
+    it "returns true if the parent is archived" do
+      parent = FactoryBot.create(:project, status: "archived")
+      sub_project1 = FactoryBot.create(:project, parent: parent)
+
+      expect(sub_project1.archived?).to eql(true)
+    end
+
+    it "returns false if the parent is not archived" do
+      parent = FactoryBot.create(:project)
+      sub_project1 = FactoryBot.create(:project, parent: parent)
+
+      expect(sub_project1.archived?).to eql(false)
+    end    
+  end
+
   describe "#toggle_archived!" do
     context "when unarchived" do
       before(:each) do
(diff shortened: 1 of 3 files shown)


bronzdoc (changes requested):
Left some comments

spec/models/project_spec.rb
```diff
@@ -21,6 +21,30 @@
     expect(sub_project2.position).to eq 2
   end
 
+  describe "#archived?" do
+    it "returns true if archived" do
+      expect(Project.new(status: "archived").archived?).to eql(true)
```
arielj, line 26: maybe we can do `expect(Project.new(status: "archived")).to be_archived`? I think it's cleaner to read and it will check that `archived?` returns `true` anyway, since by convention methods ending with `?` should return true/false I think it's ok to not be explicit on the matcher against true/false

also maybe the test name can match the other `it "returns true if status is 'archived'" do`, so on both cases we reference the status attribute

spec/models/project_spec.rb
```diff
@@ -65,6 +89,20 @@
...
     end
+
+    it "won't update a sub project" do
+      parent = FactoryBot.create(:project)
+      sub_project1 = FactoryBot.create(:project, parent: parent)
+      sub_project2 = FactoryBot.create(:project, parent: parent, status: "archived")
+
+      expect(sub_project1.toggle_archived!).to eql(nil)
+      sub_project1.reload
+      expect(sub_project1.status).to eql(nil)
```
arielj, line 100: maybe `expect(sub_project1).not_to be_archived`?

spec/models/project_spec.rb
```diff
@@ -65,6 +89,20 @@
...
+
+      expect(sub_project2.toggle_archived!).to eql(nil)
+      sub_project2.reload
+      expect(sub_project2.status).to eql("archived")
```
arielj, line 104: maybe `expect(sub_project2).to be_archived`?

spec/models/project_spec.rb
arielj: I think it's better to test the semantic than the actual attribute value since we already have tests for what each value of `status` means

arielj (commented):
hey @mcelaney I added some refactor ideas, all suggestions

mcelaney:
@arielj Thanks!  I had forgotten about the semantic sugar on booleans

mcelaney:
@bronzdoc Lemme know if you'd like to see any other changes

kindoflew:
@bronzdoc and @arielj, i think this is ready for another look. thanks!

arielj (approved):
LGTM!

rishijain:
@mcelaney hey, can you please resolve the merge conflicts?

rishijain:
I have QA this and it is working as expected. Once the merge conflicts are resolved, I

mcelaney:
@rishijain Should be good to go

rishijain:
Thank you for resolving this and for the contribution @mcelaney . :green_heart:

rishijain (approved):

PR #236, opened 2022-03-12
[WIP] Add basic setup for full text search

### LINKED ISSUE #199: Search feature for all stories across projects
https://ombulabs.atlassian.net/browse/DT-252

**Before we start...:**

- [x] I checked the documentation and didn't find this feature
- [x] I checked to make sure that this feature has not already been requested

**Branch/Commit:**

main branch

**Describe the feature:**

Please include a detailed description of the feature you are requesting and any detail on it’s expected behavior.

> **As an estimator**
> **I search for a story (by title and/or description)**
> **And then I see a bunch of stories across projects**

**Problem:**

Please include a detailed description of the problem this feature would solve.

> **As an estimator**
> **I want to see past stories**
> **So that I can understand previous estimations of similar stories**

**Challenges:**

As part of our blind estimation process, we probably don't want to display estimations for recent stories (or the project that I'm currently working on) so that we can keep estimators opinion unbiased.

mcelaney:
**WORK IN PROGRESS**

Closes #199

**Description:**

Adds a full text cross project search feature.

Sets up [pg_search](https://github.com/Casecommons/pg_search) for the Stories model and indexes search on title and description.

Screen Shot 2022-03-11 at 8 23 10 PM

___

[DT-225](https://ombulabs.atlassian.net/browse/DT-225)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -214,6 +214,9 @@ GEM
     parser (3.0.2.0)
       ast (~> 2.4.1)
     pg (1.2.3)
+    pg_search (2.3.6)
+      activerecord (>= 5.2)
+      activesupport (>= 5.2)
     public_suffix (4.0.6)
     puma (4.3.11)
       nio4r (~> 2.0)
@@ -394,6 +397,7 @@ DEPENDENCIES
   omniauth-github (~> 2.0.0)
   omniauth-rails_csrf_protection
   pg (~> 1.1)
+  pg_search
   puma (~> 4.3)
   rails (~> 6.1.4)
   rails-controller-testing
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,0 +1,9 @@
+class SearchController < ApplicationController
+  before_action :authenticate_user!
+  include ApplicationHelper
+
+  def index
+    @search_query = params.permit(:query)["query"]
+    @results = Story.search_any_word(@search_query)
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1,0 +1,303 @@
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
+SELECT pg_catalog.set_config('search_path', '', false);
+SET check_function_bodies = false;
+SET xmloption = content;
+SET client_min_messages = warning;
+SET row_security = off;
+
+SET default_tablespace = '';
+
+SET default_table_access_method = heap;
+
+--
+-- Name: ar_internal_metadata; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.ar_internal_metadata (
+    key character varying NOT NULL,
+    value character varying,
+    created_at timestamp(6) without time zone NOT NULL,
+    updated_at timestamp(6) without time zone NOT NULL
+);
+
+
+--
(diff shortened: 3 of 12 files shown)


Gemfile.lock
```diff
@@ -214,6 +214,9 @@ GEM
     parser (3.0.2.0)
       ast (~> 2.4.1)
     pg (1.2.3)
+    pg_search (2.3.6)
```
arielj, line 217: we should run `BUNDLE_GEMFILE=Gemfile.next bundle install` too so it updates the Gemfile.next.lock with this new dependency

db/structure.sql
```diff
@@ -0,0 +1,303 @@
+SET statement_timeout = 0;
```
arielj, line 1: do we need this file? is it needed because of the pg_search gem? does this affect the application setup in any way? (like we currently run `rails db:setup` I think and it looks for the schema.rb and seeds) do we need to update the README or something?

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,9 @@
+class SearchController < ApplicationController
+  before_action :authenticate_user!
+  include ApplicationHelper
+
+  def index
+    @search_query = params.permit(:query)["query"]
```
arielj, line 6: we don't need to `.permit(:query)`, we can just do `params[:query]` here

arielj (commented):
Hey, I was reviewing this and then I realized it says WIP so I'll just send the comments I had to not lose them

etagwerker:
Not sure about this, whether we want to use PG's full text search or Elasticsearch. I'd really like to know what @fbuys has to say about this.

mcelaney:
Hey @fbuys - this is a fast spike to demonstrate using pg_search - but maybe it's worth the effort to implement elasticsearch? Any thoughts?

fbuys:
I have not implemented pg_search before. I have implemented elastic search. My initial thought is that elastic search adds additional complexity and another dependency we would need to manage. If pg_search give us the result we expect I would lobby for using it over elastic search.

db/structure.sql
mcelaney: Yes - I (think I) had to [change the schema dump](https://edgeguides.rubyonrails.org/active_record_migrations.html#types-of-schema-dumps) because `searchable tsvector` doesn't seem to be supported by a Rails schema.

There is some cleanup around this I'd wanna do if we move forward with this approach - but it won't act any differently for users. This loads with the same `rails db:schema:load` because of the change I made here `[config/application.rb](https://github.com/fastruby/points/pull/236/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9)`

The upside is we get all of Postgres available to us... downside is it's less friendly to look at than a schema file to see what fields are available. Something like the [Annotate gem](https://github.com/ctran/annotate_models) might help with that.

mcelaney: (To be clear I'd wanna remove `db/schema.rb` if I move forward with this since the schema is now held by structure.sql)

mcelaney:
@etagwerker @fbuys I've done both and don't have a strong opinion... typically I'll leverage this for full text search and reach for Elastic Search if I'm doing a lot of filtering or weird sorting rules on the results.

app/controllers/search_controller.rb
mcelaney: Ah - sorry... that was me trying to get past the "turbolinks won't let me update the canvas" issue I was having.  I tried a _lot_ of stuff that didn't work :(

etagwerker:
We ended up going in a different direction. We now have an admin panel that is solving this problem (for admins only)

PR #237, opened 2022-03-13
Add extra information field to stories

### LINKED ISSUE #125: Support an `extra info` field in stories that can store context but won't be exported in the action plan
**Branch/Commit:**

Main branch

**Describe the feature:**

When we have to estimate a story that we didn't add, it's useful to have some context on what triggered that story (maybe some lines of code, the output of some grep command, etc). That information may be just accessory to help the estimation process but would be too long / irrelevant to put it in the action plan at the end.

Adding another text area to store this information can help estimators with more info while not adding noise to the final report.

**Problem:**

Sometimes, stories are missing context from the actual application code, but adding that context may be too long / irrelevant when generating the action plan.

mcelaney:
Closes #125

**Description:**

Add an extra_info field to stories which is shown to users when planning but not in the action plan itself. This allows users to add context to issues to add context details for individuals trying to  add estimates when those details won't be important after the estimating process.

Screen Shot 2022-03-12 at 9 22 03 PM
Screen Shot 2022-03-12 at 9 22 16 PM
Screen Shot 2022-03-12 at 9 22 27 PM

___

### DIFF
--- a/app/assets/javascripts/stories.js
+++ b/app/assets/javascripts/stories.js
@@ -1,18 +1,23 @@
 document.addEventListener("input", (e) => {
-  if (e.target.classList.contains("story-description")) {
-    const desc = e.target;
-    const form = desc.closest("form");
-    const preview = form.querySelector(".story_preview .content");
-    if (preview) {
-      Rails.ajax({
-        type: "POST",
-        url: "/stories/render_markdown",
-        data: `markdown=${encodeURIComponent(desc.value)}`,
-        dataType: "text",
-        success: (response) => {
-          preview.innerHTML = response;
-        },
-      });
+  function preview_setup(text_area_class, preview_class) {
+    if (e.target.classList.contains(text_area_class)) {
+      const desc = e.target;
+      const form = desc.closest("form");
+      const preview = form.querySelector(preview_class  + " .content");
+      if (preview) {
+        Rails.ajax({
+          type: "POST",
+          url: "/stories/render_markdown",
+          data: `markdown=${encodeURIComponent(desc.value)}`,
+          dataType: "text",
+          success: (response) => {
+            preview.innerHTML = response;
+          },
+        });
+      }
     }
   }
+
+  preview_setup("story-description", ".story_preview")
+  preview_setup("story-extra-info", ".extra_info_preview")
 });
--- a/app/assets/javascripts/stories.js
+++ b/app/assets/javascripts/stories.js
@@ -1,23 +1,17 @@
 document.addEventListener("input", (e) => {
-  function preview_setup(text_area_class, preview_class) {
-    if (e.target.classList.contains(text_area_class)) {
-      const desc = e.target;
-      const form = desc.closest("form");
-      const preview = form.querySelector(preview_class  + " .content");
-      if (preview) {
-        Rails.ajax({
-          type: "POST",
-          url: "/stories/render_markdown",
-          data: `markdown=${encodeURIComponent(desc.value)}`,
-          dataType: "text",
-          success: (response) => {
-            preview.innerHTML = response;
-          },
-        });
-      }
+  document.querySelectorAll('[data-has-preview]').forEach(element => {
+    const form = element.closest("form");
+    const preview = form.querySelector("." + element.dataset.previewTarget  + " .content")
+    if (preview) {
+      Rails.ajax({
+        type: "POST",
+        url: "/stories/render_markdown",
+        data: `markdown=${encodeURIComponent(element.value)}`,
+        dataType: "text",
+        success: (response) => {
+          preview.innerHTML = response;
+        },
+      });
     }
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/stories.js
```diff
@@ -1,18 +1,23 @@
...
+          success: (response) => {
+            preview.innerHTML = response;
+          },
+        });
+      }
     }
   }
+
+  preview_setup("story-description", ".story_preview")
+  preview_setup("story-extra-info", ".extra_info_preview")
```
arielj, line 22: maybe now thta we use this preview in more than one place we can do it in a more generic way?

I imagine something like using a `data-has-preview` attribute on the textareas and a `data-preview-target` attribute with the selector of the preview element, and then the javascript can be really generic like `document. querySelectorAll('[data-has-preview]').forEach(....)` so we don't have to maintain the classes both in the html and the js? what do you think?

app/assets/stylesheets/stories.scss
```diff
@@ -83,6 +88,18 @@
     }
   }
 
+  .extra_info_preview {
+    grid-area: extra-preview;
+    .content {
```
arielj, line 93: this css is duplicated with the `.content` inside `.story_preview`, we can refactor this a bit to use the same style definition

app/views/stories/_form.html.erb
```diff
@@ -28,6 +28,19 @@
     <div class="content"><%= markdown(@story.description) %></div>
   </div>
 
+  <div class="field story_extra_info">
+    <%= f.label :extra_info, "Extra Info (Markdown - for context only, not added to action plan)" %>
+    <%= f.text_area :extra_info, placeholder: "Extra Indo", class: "story-extra-info", rows: 20 %>
+  </div>
+
+  <%= f.hidden_field :project_id, value: params[:project_id] %>
```
arielj, line 36: we don't need this hidden field, it's already in line 24

spec/features/action_plan_generate_spec.rb
```diff
@@ -8,15 +8,15 @@
...
+      FactoryBot.create(:story, title: "First Story", description: "First", position: 1, project: project, extra_info: "Extra Information")
     end
   end
 
   let!(:project2) do
     FactoryBot.create(:project, parent: parent).tap do |project|
-      FactoryBot.create(:story, title: "Third Story", description: "Third", position: 2, project: project)
-      FactoryBot.create(:story, title: "Forth Story", description: "Forth", position: 1, project: project)
+      FactoryBot.create(:story, title: "Third Story", description: "Third", position: 2, project: project, extra_info: "Extra Information")
+      FactoryBot.create(:story, title: "Forth Story", description: "Forth", position: 1, project: project, extra_info: "Extra Information")
```
arielj, line 19: are these extra_info values needed? the field should be optional, so I think it's better to not add the attribute if not needed, or it will give the impression that it's required for somebody reading the tests

arielj (commented):
Hey @mcelaney , thanks! I added some comments

app/assets/javascripts/stories.js
mcelaney: @arielj  I need to admit that I'm not much of a frontend dev -- so I took a crack at this, and it works, but I'm not confident there isn't a better way to implement this.  If you have any reccos lemme know

spec/features/action_plan_generate_spec.rb
mcelaney: I left one of them in along with a comment to point out that the test asserts (by omission) the extra_info content doesn't show up in the action plan

app/assets/javascripts/stories.js
mcelaney: I spoke too soon... this seems to work in the browser but it broke the test and I can't figure out why... I might need an assist here

arielj: we can pair maybe today if you want (though now I know your complicated working hours xD)

spec/features/action_plan_generate_spec.rb
```diff
@@ -32,6 +32,7 @@
     expect(page).to have_selector("input[name='action-plan-prefix']")
     expect(page).to have_selector("h3.action-plan_heading > span", visible: false)
     expect(page.all("#action-plan h3").map(&:text)).to eq(["First Story", "Second Story"])
+    # This assertion alsp ensure the extra_info key we populated doesn't appear
```
arielj, line 35:
```suggestion
    # This assertion also ensure the extra_info key we populated doesn't appear
```

arielj (approved):
looks good to me! I ran the tests again and it passed

I added a comment for a typo but it's all good, I'll let this approved for the devtooling team to handle the next steps

rishijain:
I have reviewed and QA this as well. It looks good to move ahead. I will check internally with the team on the next steps and update here.

PR #244, opened 2022-04-06
Allow deleting a story from the story's show page

JuanVqz:
Closes #240

I will abide by de [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)

PR #245, opened 2022-04-06
Fix stories_manage_spec

fbuys:
**Description:**

Capybara failed intermittently when running spec/features/stories_manage_spec.rb. It also failed all the time when we add a sleep(1) on line 89.

Seems like the JS on this page refreshes the page after we already typed our input. This is most likely only an issue we will find with automated tests. To fix this we wait for the `turbolinks:load` event before we attach the input listener.

This seems to work for automated tests, without any regression in real world use.

This issue was found when we tried to troubleshoot another PR that kept on failing during the CI run.
https://github.com/fastruby/points/pull/244

___

### DIFF
--- a/app/assets/javascripts/stories.js
+++ b/app/assets/javascripts/stories.js
@@ -1,17 +1,21 @@
-document.addEventListener("input", (e) => {
-  document.querySelectorAll('[data-has-preview]').forEach(element => {
-    const form = element.closest("form");
-    const preview = form.querySelector("." + element.dataset.previewTarget  + " .content")
-    if (preview) {
-      Rails.ajax({
-        type: "POST",
-        url: "/stories/render_markdown",
-        data: `markdown=${encodeURIComponent(element.value)}`,
-        dataType: "text",
-        success: (response) => {
-          preview.innerHTML = response;
-        },
-      });
-    }
-  })
+document.addEventListener("turbolinks:load", function () {
+  document.addEventListener("input", (e) => {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/stories.js
```diff
@@ -1,17 +1,21 @@
...
-        url: "/stories/render_markdown",
-        data: `markdown=${encodeURIComponent(element.value)}`,
-        dataType: "text",
-        success: (response) => {
-          preview.innerHTML = response;
-        },
-      });
-    }
-  })
+document.addEventListener("turbolinks:load", function () {
```
fbuys, line 1: Only added the `turbolinks:load` listener. 
The rest is only formatting.

JuanVqz (approved):
good catch! LGTM

fbuys:
@rishijain could you please do QA and merge if it all looks good.
Essentially we need to make sure that we can edit a story and the preview shows in real time as we edit.
We also need to make sure the preview shows in real time when we create a new story.

rishijain:
On the production app, I can feel a certain delay in when I type in the description box and the preview takes about a second to load. I can feel a lag.

Why do you think there is a lag @fbuys ?

fbuys:
Hey @rishijain thanks for checking this out.
I'll take your word for it.

I found another way to fix this spec. Please see [this PR](https://github.com/fastruby/points/pull/246) and let me know what you think.

PR #246, opened 2022-04-08
Revert js changes instead add sleeps

fbuys:
**Description:**

I though the problem was related to waiting for the page to load before
adding the appropriate listeners. Instead the problem was giving the
page enough time before filling out the forms during tests.

Please include a summary of the change and which issue is fixed or which feature is introduced. If changes to the behavior are made, clearly describe what changes.
If changes to the UI are made, please include screenshots of the before and after.

Jira: https://ombulabs.atlassian.net/browse/DT-268
___

### DIFF
--- a/spec/features/stories_manage_spec.rb
+++ b/spec/features/stories_manage_spec.rb
@@ -113,6 +113,7 @@ RSpec.describe "managing stories", js: true do
 
     DESC
 
+    sleep(1)
     fill_in "story[description]", with: desc
 
     within(".story_preview .content") do
@@ -151,6 +152,7 @@ RSpec.describe "managing stories", js: true do
 
     DESC
 
+    sleep(1)
     fill_in "story[extra_info]", with: desc
 
     within(".extra_info_preview .content") do
--- a/spec/features/stories_manage_spec.rb
+++ b/spec/features/stories_manage_spec.rb
@@ -100,7 +100,7 @@ RSpec.describe "managing stories", js: true do
     end
     expect(Story.count).to eq 1
   end
-  
+
   it "shows a preview of the description while typing", js: true do
     visit project_path(id: project.id)
     click_link "Add a Story"
@@ -116,6 +116,7 @@ RSpec.describe "managing stories", js: true do
 
     expect(page).to have_text("Description Preview")
     fill_in "story[description]", with: desc
+    expect(find("#story_description").value).to have_text("This story allows users to add stories.\n\n    some\n    code\n\n")
 
     within(".story_preview .content") do
       expect(page).to have_text("This story allows users to add stories.")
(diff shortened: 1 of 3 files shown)


spec/features/stories_manage_spec.rb
```diff
@@ -113,6 +113,7 @@
 
     DESC
 
+    sleep(1)
     fill_in "story[description]", with: desc
```
JuanVqz, line 117: @fbuys if you may me, I don't think adding the sleep statement it's better to add the javascript listener, it will slow down the test suit at some point.

fbuys: That sounds reasonable @JuanVqz , however it still fails locally even with the javascript listener. Do you know if I can add a sleep for less than 1 second?

JuanVqz: I'm not pretty sure but I guess yes `sleep(0.5)` may work.

I think what about if we show a loading icon on the page when the user is writing, we can add an expectation the loading icon should not be there when the preview gets updated.

toppa: I'll share that I try to avoid `sleep` calls in test if at all possible. I'll try to have the test look for a DOM change that will indicate it's ok for the test to proceed. If for any reason that's not possible or isn't reliable, this is a module I used in a past project. It's from jQuery days... It looks like it could be adapted for non-jQuery use with this: https://stackoverflow.com/a/287212/595962

```
module WaitForAjax
  def wait_for_ajax
    Timeout.timeout(Capybara.default_max_wait_time) do
      loop until finished_all_ajax_requests?
      yield
    end
  end

  def finished_all_ajax_requests?
    page.evaluate_script('jQuery.active').zero?
  end
end

RSpec.configure do |config|
  config.include WaitForAjax, type: :feature
end
```

To give credit where it's do, I got the idea from here https://thoughtbot.com/blog/automatically-wait-for-ajax-with-capybara

I added the yield so you can use it in a block, like this:

```
wait_for_ajax do
  expect(page).to have_content('Princess Bubblegum 1 started, 0 sent')
  expect(page).to have_content('Visitors (2) Lurkers (2) Leads (1)')
end
```

fbuys: Thanks for all the feedback. It. inspired me to do more research. I found this article: https://www.varvet.com/blog/why-wait_until-was-removed-from-capybara/ and decided to use assertions that would cause capybara to wait where necessary.

toppa (approved):
Nice work! 💯

spec/features/stories_manage_spec.rb
JuanVqz: LGTM!, let's see if this solves the issue.

JuanVqz (approved):

JuanVqz:
After doing QA, it seems like the test still failing sometimes,
I just added another `expect` and it seems to be working now, I wonder if somebody else maybe @rishijain or @fbuys can do QA after my change.

JuanVqz:
it is again green! :rocket:

PR #248, opened 2022-04-29
[DT-222] Upgrade Rails to 7.0

rishijain:
This PR upgrades the app to Rails 7.0

This is in line with the work done in this pr https://github.com/fastruby/points/pull/190
Although this PR was raised to upgrade Rails to 7.0 and ruby to 3.1. Due to the ruby upgrade in this pr, the specs fail due to the matrix gem and behavior change in ruby 3.1.

I think we should keep this as 2 different tasks as the app is currently on ruby 3.0 which also works with rails 7.0.
We can take up the ruby upgrade work separately and I don't believe it will be a lot.

Lets test Rails upgrade on staging before we merge this in.

https://ombulabs.atlassian.net/browse/DT-222


Gemfile
```diff
@@ -86,8 +86,7 @@ group :development do
   gem "web-console", ">= 3.3.0"
   gem "listen", "~> 3.7"
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
-  gem "spring"
-  gem "spring-watcher-listen", "~> 2.0.0"
```
rishijain, line 90: spring needs to be upgraded to 3.0.0 with Rails 7.0 and the gem `spring-watcher-listen` latest version 2.0.1 depends on spring < 3.0 

So to make this work, we need to remove this gem
refer https://rubygems.org/gems/spring-watcher-listen/versions/2.0.0

rishijain:
I am closing this pr to upgrade Rails 7.0 and ruby to 3.1 https://github.com/fastruby/points/pull/190 and have raised a new issue for updating ruby from 3.0 to 3.1 https://github.com/fastruby/points/issues/249

rishijain:
hey @fbuys @JuanVqz , can one of you please review this PR, and once this is ready, let's deploy this to staging and test the branch there before the merge to main. Thanks.

Gemfile
arielj: hmmm I wonder why this change was not needed on the dual boot with the main branch before this change (since the main branch should be rails 7), maybe the dual boot against `main` is not setup properly?

app/controllers/projects_controller.rb
```diff
@@ -56,7 +56,7 @@ def create
       redirect_to "/projects"
     else
       flash[:error] = @project.errors.full_messages
-      redirect_back(fallback_location: "projects/new")
+      render :new
```
arielj, line 59: you should probably do `flash.now[:error] = ...` if you render instead of redirect

Gemfile
```diff
@@ -15,7 +15,7 @@ if next?
   gem "rails", github: "rails/rails", branch: "main"
   gem "devise", github: "heartcombo/devise", branch: "main"
 else
-  gem "rails", "~> 6.1.4"
+  gem "rails", "~> 7.0.2.4"
```
arielj, line 18: maybe use `~> 7.0.2` here? the current version will only allow `7.0.2.5/6/7/8..` but not `7.0.3` for example, and we usually limit the minor version but not the patch version

rishijain:
hey @arielj can you please have a further look at this pr?

rishijain:
ping @arielj

kindoflew:
ping @rishijain -- did you get a chance to look at Ariel's new(ish) comment?: https://github.com/fastruby/points/pull/248#discussion_r871294257

thanks! :shipit:

app/controllers/projects_controller.rb
fbuys: @rishijain did you see this comment from Ariel?

fbuys (commented):
Hey @rishijain it looks great, I asked some questions - hope they don't sound too ignorant 🙈

Gemfile
rishijain: Hey @fbuys I have verified the project is working when booted with next rails server.

app/controllers/projects_controller.rb
JuanVqz: @arielj are you happy with this [answer](https://github.com/fastruby/points/pull/248#discussion_r872256868)?

config/application.rb
```diff
@@ -9,7 +9,7 @@
 module Points
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+    config.load_defaults 7.0
```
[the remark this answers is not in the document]
JuanVqz: Aside note: if we were moving from `config.load_defaults 5.1` to `config.load_defaults 6.0` we could decide if go with the default behavior which adds `zeitwerk` of disable `zeitwerk` adding the following both lines:

```
config.load_defaults 6.0
config.autoloader = :classic
```
but since we moved right to `config.load_defaults 7.0` there are no more options regarding `autoloading`, `zeitwerk` is a must.

JuanVqz:
The code looks good to me

@rishijain, thank you for contributing :1st_place_medal:  

I'm assigning to @fbuys to do QA

JuanVqz (approved):

fbuys (changes requested):
I ran: `BUNDLE_GEMFILE=Gemfile.next rspec` and saw:
image

Without next:
image

rishijain:
Cool. will check.

I did this locally `next bundle exec rspec` and everything passed, so it seemed to me nothing is breaking. I will try to fix this. Thanks for the check @fbuys  :)

JuanVqz:
> I ran: `BUNDLE_GEMFILE=Gemfile.next rspec` and saw: image
> 
> Without next: image

did you run it several times? isn't flaky tests?

PR #254, opened 2022-05-12
[DT-262] Fix button colors are not consistent

### LINKED ISSUE #99: Button colors are not consistent
https://ombulabs.atlassian.net/browse/DT-262

**Branch/Commit:**

Main branch

**Expected behavior:**

The color of the buttons with similar actions should be consistent. Buttons deleting stuff should be red-ish, buttons adding stuff should be green, other buttons should be yellow (just an example, not sure if those are the right decisions)

**Actual behavior:**

The colors are not consistent:
- we have a red button to bulk delete, but a green button to delete a singles story
- we have a red button to add a story, but a green button do add an estimate
- we have a green button to edit a story or an estimate, but a red button to edit the project
- we have a green button to edit a story, but the submit button is yellow, and the back button is yellow too

**Screenshots and Videos**

image
image

## EDIT:

After checking with nathalie, she gave some suggestions:
- the row actions are changing to something like this for now
image

but she proposed something like this for the future:
image
image

- the `Bulk Delete` button:
- - would be better if it looks like a link instead of a button so it doesn't look that important, (or maybe it shouldn't be visible until stories are selected? added by me)
- - the `Bulk Delete` button can be renamed to `Delete`, the `x story/ies` makes it clear it can be a bulk action
- - (added by me), maybe move bulk actions into a `more` dropdown? we could have bulk move / delete / clone for example

- the `Return to Projects Page` button should be a button, and shouldn't be there with the actions, it's a navigation action so it should be in the header or the nav bar, and should look like a button but like a link since it's navigation

JuanVqz:
Closes #99 

- [x] Align back button to the left (I think it's more user friendly)
- [x] [Change button's color following Ariel's convention](https://github.com/fastruby/points/issues/99#issuecomment-957473643)

[DT-262](https://ombulabs.atlassian.net/browse/DT-262)

image
image
image
DT-262

___

### DIFF
--- a/app/assets/stylesheets/4-molecules/_floors.scss
+++ b/app/assets/stylesheets/4-molecules/_floors.scss
@@ -13,7 +13,6 @@
 		justify-content: center;
 		width: 100%;
 		overflow: hidden;
-		background-color: #D9F1F1;
 		height: calc(100vh - 160px);
 		margin: -100px 0;
 	}
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -7,7 +7,7 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
-    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), class: "button", id:"back", method: :post %><br />
+    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), class: "button green", id:"back", method: :post %><br />
   <% end -%>
 <% end -%>
 </div>
--- a/app/views/estimates/_form.html.erb
+++ b/app/views/estimates/_form.html.erb
@@ -26,7 +26,7 @@
   </div>
 
   <div class="btn-group">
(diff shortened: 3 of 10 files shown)


app/assets/stylesheets/4-molecules/_floors.scss
```diff
@@ -13,8 +13,7 @@
 		justify-content: center;
 		width: 100%;
 		overflow: hidden;
-		background-color: #D9F1F1;
```
JuanVqz, line 16: I removed the background because 2 things:
1. I feel with white background is more consistent with other views
2. I feel like green is not matching with it. (I'm not an expert on colors.)

image

image

arielj:
When I added the comments on that issue, I didn't know nathalie was allowed to dedicate time to the open source projects, I think it's a good idea to pass these changes through her first for validation (I added some ideas but it was just my opinion).

rishijain:
I agree with @arielj on letting Nathalie also having a look at these.
Btw, I did a QA on this and it is working fine @JuanVqz .:)

etagwerker:
@JuanVqz This looks good to me. Could you please address the conflicts? Thanks!

JuanVqz:
> @JuanVqz This looks good to me. Could you please address the conflicts? Thanks!

@etagwerker conflicts are addressed. 🚀

etagwerker (approved):

PR #262, opened 2022-06-15
Enable/disable RMP with a cookie, README

arielj:
**Description:**

This PR changes the rack mini profiler setup to allow the user to enable/disable passing an `?enable_rack_mini_profiler` or `?disable_rack_mini_profiler` query estring parameter.

The query string param also applies for the development environment.

With this we don't need the Ctrl+Alt+P shortcut to toggle it on/off after it rendered and we can disable it completely to avoid extra requests and extra code in responses unless manually enabled.

___

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
-  before_action :enable_rack_mini_profiler
+  before_action :rack_mini_profiler
 
   def hello
     render html: "hello, world!"
@@ -13,7 +13,18 @@ class ApplicationController < ActionController::Base
   private
 
   # we can enable this for everybody, it's hidden by default
-  def enable_rack_mini_profiler
-    Rack::MiniProfiler.authorize_request
+  def rack_mini_profiler
+    show =
+      if params.key?(:enable_rack_mini_profiler)
+        cookies[:rack_mini_profiler] = "show"
+        true
+      elsif params.key?(:disable_rack_mini_profiler)
+        cookies.delete(:rack_mini_profiler)
+        false
+      end
+
+    if show || cookies[:rack_mini_profiler]
+      Rack::MiniProfiler.authorize_request
+    end
   end
 end
--- a/config/initializers/rack_mini_profiler.rb
+++ b/config/initializers/rack_mini_profiler.rb
@@ -1,2 +1,1 @@
-Rack::MiniProfiler.config.start_hidden = true
-Rack::MiniProfiler.config.toggle_shortcut = "Alt+Ctrl+P"
+Rack::MiniProfiler.config.authorization_mode = :allow_authorized
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ and then:
 User.find_by(email: "user@example.com").update_attribute(:admin, true)
 ```
 
+## Rack Mini Profiler
+
+To enable the Rack Mini Profiler widget, add the `?enable_rack_mini_profiler` query string to any url. The widget will stay on for all requests until disabled.
+
+To disable the Rack Mini Profiler widget, add the `?disable_rack_mini_profiler` query string to any url.
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at [https://github.com/fastruby/points](https://github.com/fastruby/points). This project is intended to be a safe, welcoming space for collaboration, and contributors are expected to adhere to the [Contributor Covenant](http://contributor-covenant.org) code of conduct.


app/controllers/application_controller.rb
```diff
@@ -13,7 +13,18 @@ def after_sign_in_path_for(resource)
...
+        cookies[:rack_mini_profiler] = "show"
+        true
+      elsif params.key?(:disable_rack_mini_profiler)
+        cookies.delete(:rack_mini_profiler)
+        false
+      end
+
+    if show || cookies[:rack_mini_profiler]
+      Rack::MiniProfiler.authorize_request
+    end
```
arielj, line 28: I tried doing a one-liner here but sometimes the widget didn't show up, no idea what was going on there, so I used a normal if / end

config/initializers/rack_mini_profiler.rb
```diff
@@ -1,2 +1 @@
-Rack::MiniProfiler.config.start_hidden = true
-Rack::MiniProfiler.config.toggle_shortcut = "Alt+Ctrl+P"
\ No newline at end of file
+Rack::MiniProfiler.config.authorization_mode = :allow_authorized
```
arielj, line 1: so the query string switch applies to development too

fbuys (approved):
One minor comment, but it works great!

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
-  before_action :enable_rack_mini_profiler
+  before_action :rack_mini_profiler
```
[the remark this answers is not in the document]
arielj: hmmm I wasn't sure, since it's not only toggling the thing, but I'm ok with it, I don't think there's a word for `toggle and maybe show it if needed` xD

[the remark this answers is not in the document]
fbuys: I had to refresh the page after it was toggled on

app/controllers/application_controller.rb
```diff
@@ -13,7 +13,18 @@ def after_sign_in_path_for(resource)
...
-    Rack::MiniProfiler.authorize_request
+  def rack_mini_profiler
+    show =
+      if params.key?(:enable_rack_mini_profiler)
+        cookies[:rack_mini_profiler] = "show"
+        true
+      elsif params.key?(:disable_rack_mini_profiler)
+        cookies.delete(:rack_mini_profiler)
+        false
+      end
```
JuanVqz, line 24: @arielj I don't want to be a blocker, you are free to pick this suggestion.

I'm seen that we are adding `enable_rack_mini_profiler` and `disable_rack_mini_profiler` but I think it will be a little bit better if we use it as `/projects?rack_mini_profile=true` or `/projects?rack_mini_profile=false`.
that way we have consistence in the argument and will be easy to just change `true/false` to enable/disabled.

JuanVqz (approved):
I left a suggestion, the code seems good to me :+1:

JuanVqz:
just following this up @arielj could you please fix merge conflicts? and also could you please respond to my suggestion [here](https://github.com/fastruby/points/pull/262#discussion_r905379272)?

app/controllers/application_controller.rb
arielj: hmmm I wasn't sure about this, should we only support `true`/`false`? what about `1`/`0`? or `True`/`False`? or `enable`/`disable`? or `enabled`/`disabled`?

I think that calling the query string param `enable_rack_mini_profiler` and `disable_rack_mini_profiler` is more expressive and explicit, you either enable or disable rack_mini_profiler.

Using `rack_mini_profiler=true` or `=false` leaves me with some questions, what does it do? it enables it? for how long? why is it `true`/`false` and not `enable`/`disable`? I feel it's more `developer-ish` but not that developer friendly.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
arielj: hmmm I think something might be happening the very frist time it's toggled in development after the server was started, that's the only consistent way I can reproduce this:

- make sure RMP is disabled, no cookie
- start the server `rails s`
- open any page
- add `?enable_rack_mini_profiler` and enter
- the widget doesn't show
- after this, disabling/enabling RMP shows it just fine every time without that extra refresh

I'm debugging this and the call to `authorize_request` is done correctly, for some reason it's not working that very first time, and I suspect something with development environment and lazy loading of classes? I don't think it's worth to do much more research, it seems to be something on the RMP gem

arielj:
@JuanVqz @fbuys I fixed the conflicts and also replied to your comments

JuanVqz:
@fbuys LGTM :+1:

etagwerker:
@fbuys If it all looks good, would you please merge this PR? Thanks!

fbuys (approved):

PR #178, opened 2022-01-06
Prevent changes to archived project

cleicar:
**Description:**

closes #154 

This PR disables the action buttons when the project is archived and redirects to the projects page when trying to force the changes.

Screen Shot 2022-01-06 at 18 54 33

___

I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/pull_request_template.md).

jira link https://ombulabs.atlassian.net/browse/DT-223

PR #263, opened 2022-06-16
Prevent changes to an archived project

### LINKED ISSUE #154: Archived projects shouldn't allow modifications
https://ombulabs.atlassian.net/browse/DT-256

**Branch/Commit:**

Main branch.

**Expected behavior:**

When a project is archived, users shouldn't be able to change stories or estimates. Those actions should be disabled and trying to force any of them should redirect back to the project's show page with an error saying that you can't modify an archived project.

This should also apply to changing things in sub-projects of an archived parent project.

**Actual behavior:**

Currently, if a project is archived, you can still go and change it (update stories, estimates, change the project title, add sub-projects, modify the sub-projects, etc).

cleicar:
**Description:**

closes #154 

Previous PR https://github.com/fastruby/points/pull/178

This PR disables the action buttons when the project is archived and redirects to the projects page when trying to force the changes.

Screen Shot 2022-01-06 at 18 54 33

___

jira link https://ombulabs.atlassian.net/browse/DT-223


app/helpers/projects_helper.rb
```diff
@@ -3,4 +3,23 @@ def estimated(story)
...
+
+  private
+
+  def text_content(icon, text)
+    if icon
+      icon_tag = content_tag(:i, "", class: "fa fa-#{icon}").concat(content_tag(:span, text))
+    end
+
+    icon_tag || text
+  end
```
JuanVqz, line 24: Same here

```suggestion
  def text_content(icon, text)
    return text if icon.nil?
    
    content_tag(:i, "", class: "fa fa-#{icon}").concat(content_tag(:span, text))
  end
```

app/helpers/projects_helper.rb
```diff
@@ -3,4 +3,23 @@ def estimated(story)
...
   end
+
+  def link_unless_archived(project, text, url, classes: nil, method: :get, remote: false, icon: nil, data_attr: nil, id: nil)
+    link_text = text_content(icon, text)
+    if project.archived? || project.parent&.archived?
+      button_tag(link_text, disabled: true, class: "button #{classes} disabled", aria: {disabled: true})
+    else
+      link_to(link_text, url, class: "button #{classes}", title: text, method: method, remote: remote, data: data_attr, id: id)
+    end
+  end
```
JuanVqz, line 14: This may be a personal way to do things in this scenario, so, maybe just for the record.

Pros:
  - you can pass more data if desired.
Cons:
  - many changes, again (and we may don't want that).
  - need to test if what I put as example works as expected.

```suggestion
  def link_unless_archived(project, text, url, clases: nil, options = { method: :get, remote: false })
    link_text = text_content(icon, text)
    if project.archived? || project.parent&.archived?
      button_tag(link_text, { disabled: true, aria: disabled: true, class: "button #{classes} disabled" }.merge(options))
    else
      link_to(link_text, url, { class: "button #{classes}" }.merge(options))
    end
  end
```

JuanVqz (commented):
I left some comments, but not pretty sure if Cleiviane should fix all my comments.
what are your thoughts about that?

app/assets/stylesheets/project.scss
```diff
@@ -109,6 +109,10 @@
     background: white;
     box-shadow: 1px 1px 4px 2px rgba(0, 0, 0, 0.4);
 
+    a span{
```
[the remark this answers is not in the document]
fbuys: Thanks, I made this change :-)

app/helpers/projects_helper.rb
fbuys: Nice one, I updated the code.

app/helpers/projects_helper.rb
fbuys: I like this suggestion @JuanVqz, I do think it is a lot of changes - so not worth our time at this stage?

fbuys (approved):
Works as expected, thank you very much @cleicar !
Also thank you for the review @JuanVqz

PR #264, opened 2022-06-23
DT-372 A PM should be able to lock a project

aisayo:
https://ombulabs.atlassian.net/browse/DT-327

- [x] A PM can lock a project
- [x] Automated tests are added to confirm this feature works

### ⚠️ Remember to migrate manually after merge


app/views/projects/show.html.erb
```diff
@@ -34,11 +34,12 @@
                 <td class="project-table__cell"><%= story.estimate_for(current_user)&.best_case_points %></td>
                 <td class="project-table__cell"><%= story.estimate_for(current_user)&.worst_case_points %></td>
                 <td class="project-table__cell story_actions">
+                <% if is_locked?(@project) %>
                   <% if estimated(story) %>
                     <%= link_unless_archived(@project, "Edit Estimate", edit_project_story_estimate_path(@project.id, story, @estimate_id), classes: "edit-estimate", remote: true) %>
                   <% else %>
                     <%= link_unless_archived(@project, "Add Estimate", new_project_story_estimate_path(@project.id, story), classes: "add-estimate", remote: true) %>
-                  <% end %>
+                  <% end %>
```
fbuys, line 42: Let's remove the whitespace at the end of this line :-)

app/helpers/projects_helper.rb
```diff
@@ -4,6 +4,10 @@ def estimated(story)
     @estimate_id.present?
   end
 
+  def is_locked?(model)
```
fbuys, line 7: Instead of using a helper could we use:
```
<% if policy(@post).update? %>
```
or the modal directly?
```
project.locked?
```

fbuys: Here is one or the reasons I prefer to avoid helpers: https://thepugautomatic.com/2012/08/helpers/

db/migrate/20220621141342_add_locked_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddLockedToProjects < ActiveRecord::Migration[6.1]
+  def change
+    add_column :projects, :locked, :datetime
```
fbuys, line 3: Could we rename this to locked_at, since it is a datetime field?

spec/controllers/projects_controller_spec.rb
```diff
@@ -158,6 +158,20 @@
...
   end
 
+  describe "#toggle_locked" do
+    context "when locking a project" do
+      it "returns a datetime" do
+        patch :toggle_locked, params: {id: project.id}, xhr: true
+        expect(assigns[:project]).to be_locked
+      end
+
+      it "returns a js response" do
```
fbuys, line 168: Why is it important that we test the response type?

spec/controllers/stories_controller_spec.rb
```diff
@@ -110,7 +110,7 @@
         expect(assigns(:story)).to eq story
       end
     end
-
+
```
fbuys, line 113: Some whitespace 🔥

spec/controllers/stories_controller_spec.rb
```diff
@@ -120,13 +120,12 @@
         expect(story.reload.title).to eq "New Story"
       end
     end
-
+
```
fbuys, line 123: 🔥

spec/features/projects_manage_spec.rb
```diff
@@ -330,4 +330,29 @@
...
+    it "locks a project" do
+      visit project_path(id: project.id)
+      expect(page).to have_selector(:link_or_button, "Lock Project")
+      click_button "Lock Project"
+      ["Delete Project", "Lock Project", "Add Sub-Project", "Add a Story"].each do |btn|
+        expect(page).not_to have_selector(:link_or_button, btn)
+      end
+    end
+
+    before do
```
fbuys, line 347: Is this required?

spec/features/projects_manage_spec.rb
```diff
@@ -330,4 +330,29 @@
...
+    end
+
+    it "locks project stories" do
```
fbuys, line 350: Perhaps we can do:
```
context "when a project is locked' do
  it "hides edit and delete buttons" do
```

Hope I understand the purpose of this test correctly?

spec/features/stories_manage_spec.rb
```diff
@@ -269,7 +269,7 @@
   end
 
   it "allows sorting stories", js: true do
-    FactoryBot.create(:story, project: project)
+    FactoryBot.create(:story, project: project, title: "Juan and Aysan Code!")
```
fbuys, line 272: Nice! 🌟

fbuys (changes requested):
Hey, I left a few comments.
Feel free to let me know where you don't agree with anything I mentioned :-)

app/helpers/projects_helper.rb
JuanVqz: Hey @fbuys, we will update it as you requested.

But I'd like to mention that I disagreed a little bit, I truly believe that the best way is doing things like the post explains but how often do we disabled global helpers in rails? rails will load helpers anyway.

Also having in as `<% if policy(@post).update? %>` will tend to add more stuff there in the future because will be the easy way `<% if policy(@post).update? && foo && bar %>`

spec/features/stories_manage_spec.rb
aisayo: Team work!

spec/controllers/projects_controller_spec.rb
JuanVqz: if someone wants to hit this endpoint with an html or json request, they will notice that we are only responding within `js` format for now. what do you think?

fbuys: Good point 👍

spec/controllers/stories_controller_spec.rb
```diff
@@ -120,13 +120,12 @@
...
+ 
     describe "#bulk_destroy" do
+     let(:stories) { FactoryBot.create_list(:story, 2, project: project) }
+
       it "deletes multiple stories" do
-        ids = [99, 100]
-        ids.each do |id|
-          FactoryBot.create(:story, id: id, project: project)
-        end
+        ids = stories.map(&:id)
```
JuanVqz, line 128: I'd like to mention that we fix this flaky test

spec/features/stories_manage_spec.rb
JuanVqz: This was a flaky test, we just fixed it

app/helpers/projects_helper.rb
aisayo: We decided to keep this helper for the purpose of refactoring and encapsulating the logic in case of needing to update anything, it is easier to refer to this one method than every where this helper is used.

spec/factories/projects.rb
```diff
@@ -1,5 +1,6 @@
 FactoryBot.define do
   factory :project do
     title { Faker::Company.name }
+    locked_at { nil }
```
fbuys, line 4: It is null by default, I don't think we need to define it in the factory.

app/views/projects/show.html.erb
```diff
@@ -96,11 +98,21 @@
...
 
   <div class="btn-group actions">
-    <%= link_unless_archived(@project, "Delete Project", project_path(@project.id), classes: "delete", method: :delete, remote: true, data_attr: { confirm: 'Are you sure?' }, id: "delete")%>
-    <%= link_unless_archived(@project, "Add Sub-Project", project_new_sub_project_path(@project), classes: "magenta") unless @project.parent_id %>
+    <% if is_locked?(@project) && current_user.admin? %>
+      <%= button_to 'Lock Project', toggle_locked_project_path(@project.id), method: "patch", class: "button magenta ", id: "lock-btn", remote: true %>
+    <% end %>
+
+    <% if is_locked?(@project) %>
+      <%= link_unless_archived(@project, "Delete Project", project_path(@project.id), classes: "delete", method: :delete, remote: true, data_attr: { confirm: 'Are you sure?' }, id: "delete")%>
```
fbuys, line 109: We should still be able to archive a project even if it is not locked

fbuys (changes requested):
Looks really great! 

I think there is one last issue left.
image

It does not look like sub-projects are locked when their parent projects are locked.

JuanVqz (approved):
I paired with @aisayo, so, I'm approving it :1st_place_medal:

fbuys:
I pulled all the latest changes.
But when I try to visit an existing project (one that existed before this branch). I see an error:
image

fbuys:
I think I need to rollback and migrate again

JuanVqz:
> I think I need to rollback and migrate again


did this solve the issue?

fbuys:
> > I think I need to rollback and migrate again
> 
> did this solve the issue?

Yes, sorry about the false alarm

fbuys:
> Looks really great!
> 
> I think there is one last issue left. image
> 
> It does not look like sub-projects are locked when their parent projects are locked.

Testing locally I created a project with a sub-project. When I lock the parent project, it seems like the sub-project is not locked. So I think we need to add something that locks sub projects when their parent projects are locked.

JuanVqz:
good catch!  both are fixed now :+1:

spec/policies/project_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(subject).not_to permit(user, locked_project)
+      end
+
+      it "grants access if project is not locked" do
+        expect(subject).to permit(user, project)
+      end
+    end
+
+    context "when sub project belongs to a locked parent" do
+      let(:sub_project) { FactoryBot.create(:project, locked_at: nil, parent: locked_project) }
```
fbuys, line 22: Think we can use a stub here? Or do we need to `create` the sub_project?

spec/policies/project_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      it "denies access" do
+        expect(subject).not_to permit(user, sub_project)
+      end
+    end
+
+    context "when sub project belongs to an unlocked parent" do
+      let(:sub_project) { FactoryBot.create(:project, locked_at: nil, parent: project) }
```
fbuys, line 30: And here `build` or `build_stubbed`

fbuys (changes requested):
I think the last thing I just picked up is that you can still bulk delete stories on a locked project. 
image

image

spec/policies/project_policy_spec.rb
JuanVqz: That's true! good catch!

spec/policies/project_policy_spec.rb
JuanVqz: > Great job so far!! tada
> 
> I think the last thing I just picked up is that you can still bulk delete stories on a locked project. image
> 
> image

Again thank you for pointing this out, fixed

JuanVqz:
> Great job so far!! tada
> 
> I think the last thing I just picked up is that you can still bulk delete stories on a locked project. image
> 
> image

@fbuys addressed!

fbuys:
Great job @JuanVqz  and @aisayo

fbuys (approved):

fbuys:
Migrated on Heroku: 
image

PR #269, opened 2022-07-14
Ruby upgrade to 3.1.0

rishijain:
**Description:**

Jira link: https://ombulabs.atlassian.net/browse/DT-378

Ruby is upgraded to 3.1.0.

Changes:
Version changed in `Gemfile`
Version updated in `.github/workflows/tests.yml`.

___


Gemfile
```diff
@@ -32,6 +32,7 @@ gem "sass-rails", "~> 5.0"
 gem "uglifier", ">= 1.3.0"
 
 gem "bourbon"
+gem "matrix"
 
 # See https://github.com/rails/execjs#readme for more supported runtimes
```
rishijain, line 37: Need to add this gem in ruby 3.1 explicitly.

Read more about it here: https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/
and an stackoverflow answer https://stackoverflow.com/questions/71991514/rails-require-cannot-load-such-file-matrix

fbuys (approved):

PR #273, opened 2022-07-29
DT-328 unlock a project

aisayo:
Added 'Unlock Project' button to project show page to support https://ombulabs.atlassian.net/browse/DT-328
Includes toggle behavior in ProjectsController in toggle_locked which will update a projects locked_at attribute on click
Added feature tests to support unlock permissions only restricted to the admin user.

These changes will complete the admin user lock feature for Points by allowing only an admin to lock or unlock a project. 

# Testing notes
- Log in with an admin user (make your user an admin using the console rails c to `user.update(admin: true)`
- View an project (or create one if you don't have any existing projects locally)
- Click on the lock button to lock the project
- Confirm that the `lock button` is replaced with `unlock button`
- Click 'Unlock Project' to toggle the project from locked to unlocked. 
- Confirm that the project is unlocked. (The locked label should disappear and edit buttons should show again)

Screen Shot 2022-07-29 at 11 50 10 AM

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -37,7 +37,11 @@ class ProjectsController < ApplicationController
   # PATCH /projects/1/toggle_locked.js
   def toggle_locked
     @project = Project.find(params[:id])
-    @project.update(locked_at: Time.current)
+    if @project.locked_at.nil?
+      @project.update(locked_at: Time.current)
+    else
+      @project.update(locked_at: nil)
+    end
   end
 
   def new_clone
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,6 +1,6 @@
 class ProjectsController < ApplicationController
   before_action :authenticate_user!
-  before_action :find_project, only: [:show, :edit, :update, :sort, :sort_stories, :destroy, :new_sub_project]
+  before_action :find_project, only: [:show, :edit, :update, :sort, :sort_stories, :destroy, :new_sub_project, :toggle_archive,:toggle_locked]
   before_action :ensure_unarchived!, only: [:edit, :new_sub_project, :update]
 
   def index
@@ -30,13 +30,11 @@ class ProjectsController < ApplicationController
   end
 
   def toggle_archive
-    @project = Project.find(params[:id])
     @project.toggle_archived!
   end
 
   # PATCH /projects/1/toggle_locked.js
   def toggle_locked
-    @project = Project.find(params[:id])
     if @project.locked_at.nil?
       @project.update(locked_at: Time.current)
     else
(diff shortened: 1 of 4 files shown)


app/controllers/projects_controller.rb
```diff
@@ -37,7 +37,11 @@ def toggle_archive
   # PATCH /projects/1/toggle_locked.js
   def toggle_locked
     @project = Project.find(params[:id])
```
fbuys, line 39: One quick improvement is to make use of the `find_project` before action instead of duplicating code here.

spec/features/projects_manage_spec.rb
```diff
@@ -373,6 +372,12 @@
 
           expect_buttons_to_be_hidden
         end
+
+        it "does not render unlocked button" do
+          visit project_path(id: project.id)
+
```
fbuys, line 378: As a personal preference, I often remove these empty lines from small tests like this one. Not a blocker at all, just wanted to share that.

spec/features/projects_manage_spec.rb
```diff
@@ -427,9 +432,39 @@
...
+        let(:locked_project) { FactoryBot.create(:project, :locked) }
+
+        it "shows an unlocked button" do
+          visit project_path(locked_project)
+          within "div.btn-group.actions" do
+            expect(page).to have_text("Unlock Project")
+          end
+        end
+
+        it "unlocks a project" do
```
fbuys, line 449: Perhaps we can be a bit more descriptive about the action that leads to the project being unlocked?
```suggestion
        it "unlocks the project when 'Unlock Project' is clicked" do
```

spec/features/projects_manage_spec.rb
```diff
@@ -427,9 +432,39 @@
...
+          end
+        end
+      end
+    end
+  end
+
   def expect_buttons_to_be_hidden
     ["Delete Project", "Lock Project", "Add Sub-Project", "Add a Story"].each do |btn|
-      expect(page).not_to have_selector(:link_or_button, btn)
+      within "div.btn-group.actions" do
```
fbuys, line 465: I think this is a risky change because I am not sure all buttons are within the btn-group div? Or if we add buttons in the future outside this div, then this test will not fail even if it should.

fbuys (commented):
Looks pretty good, nice!
I made some smaller notes, let me know what you think.

Please add testing instructions / notes for QA in the PR description.

spec/features/projects_manage_spec.rb
aisayo: Currently the buttons are within that div but I can see how that would be problematic. Would moving the scope up to the div.container be more optimal?

fbuys (approved):

torresga (approved):

PR #279, opened 2022-10-18
Testing Ombulabs Auth gem

tiagocasanovapt:
### Motivation / Context

> _Refactor a rack middleware (probably a ruby gem) out of https://points.ombulabs.com/ so that we can easily add GitHub-based authentication to other Ruby/Rails internal applications._

More @ https://github.com/fastruby/ombu_labs-auth/issues/1
    
### QA / Testing Instructions
- Login :octocat: 

___


tiagocasanovapt:
from https://github.com/fastruby/ombu_labs-auth/issues/4#issuecomment-1284185925

> _Working as a mounted engine with controllers + models + views, but.... 🧐_
>
> _....still 1 bug remaining at [db26656](https://github.com/fastruby/points/pull/279/commits), the command sign_in_and_redirect @user is trying to use the root route from the main app which is not recognized by the current Devise context_

tiagocasanovapt:
Working with the following change 😎 - https://github.com/fastruby/ombu_labs-auth/issues/4#issuecomment-1284394700

etagwerker:
Closing in favor of #297

PR #283, opened 2022-12-19
fix create button behavior on estimating

### LINKED ISSUE #266: The Create button gets odd when estimating
While working on DT-372 I found this error.

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

**Expected behavior:**

Be able to create an estimation.

**Actual behavior:**

The button to save estimations gets odd.

error-points

DamonClark:
Closes #266 

Adds value attribute to the submit button in views.

image

___

### DIFF
--- a/app/views/estimates/_form.html.erb
+++ b/app/views/estimates/_form.html.erb
@@ -26,7 +26,7 @@
   </div>
 
   <div class="btn-group">
-    <%= f.submit yield(:button_text), class: "button", id: "edit" %>
+    <%= f.submit yield(:button_text), value: "Create", class: "button", id: "edit" %>
   </div>
 
   <% if estimate.persisted? %>
--- a/spec/features/estimates_manage_spec.rb
+++ b/spec/features/estimates_manage_spec.rb
@@ -52,7 +52,7 @@ RSpec.describe "managing estimates", js: true do
 
     it "allows me to edit an estimate" do
       set_estimates(1, 2)
-      click_button "Save Changes"
+      click_button "Create"
 
       expect(page).to have_content "1"
       expect(page).to have_content "2"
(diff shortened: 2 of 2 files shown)


JuanVqz:
Hey, @DamonClark thanks for taking the time to solve this issue, 
I ran the test suite twice but [still failed,](https://github.com/fastruby/points/actions/runs/3727519372/jobs/6543246785).
Could you please check why the tests are failing? 
If you have any questions, feel free to ask we can always help.

Happy new year :+1:

DamonClark:
Hey @JuanVqz  looks like they are failing because the buttons isn't "Save Changes" but "Create" I'll commit the change to fix

JuanVqz:
> Hey @JuanVqz looks like they are failing because the buttons isn't "Save Changes" but "Create" I'll commit the change to fix.
> 
> Happy new year !

Good call, Thanks for fixing them but now I'm concerned if "Save Changes" is better wording for this action, let me get a second point of view, sorry about that.

Hey, @fbuys | @arielj  what do you think about this?
In the beginning, I agreed with calling the button "Create" because is the way it shows up the first time when you click the "Add estimate" button but what if you are updating the estimations the next time, it makes me think "Save Changes" it's better wording.

aisayo:
Hi, I think using `save changes` for this button makes the most sense since even when we are entering estimates for the first time, we are simply updating the points from 0 to an actual value. Makes more sense than using 'create'. What do you think @fbuys ?

etagwerker:
@aisayo I agree with this: 

> Hi, I think using save changes for this button makes the most sense since even when we are entering estimates for the first time, we are simply updating the points from 0 to an actual value. Makes more sense than using 'create'.

etagwerker:
@DamonClark Thanks for getting this started! I took some of your commits and continued this PR with #288 so I'm going to close this PR in favor of that one.

PR #293, opened 2023-03-30
Upgrade to Puma 6.2

etagwerker:
### Jira Ticket 

https://ombulabs.atlassian.net/browse/DT-430

### Motivation / Context

We should try to use the latest version of Puma. 
    
### QA / Testing Instructions

Manually tested that things work in development environment. 

### Screenshots:

N/A

___


fbuys (approved):
Looks great, I tested it using heroku review app and it works.

Issue #296, opened 2023-05-09
Better tracking for version jumps and projects


etagwerker:
**Before we start...:**

- [x] I checked the documentation and didn't find this feature
- [x] I checked to make sure that this feature has not already been requested

**Branch/Commit:**

`main` branch. 

**Describe the feature:**

> **As an admin**
> **I create a project**
> **And then I select what version jump it's connected to**
> **And I see version jumps in the admin panel**

**Problem:**

> **As an admin**
> **I want to be able to filter all the Rails 5.2 to 6.0 project**
> **So that I can achieve better gain insights from past projects related to that jump**

**Information Architecture:**

We probably want to have a new table called "jumps" or "version_jumps" that has these columns: 

- Technology: Ruby/Rails/Node/React/React Native/Expressjs/Vuejs
- Current version: "2.3" (for example) 
- Target version: "3.0" (for example)

When creating a new project, I should be able to select a specific jump associated with the project. 

**I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)**

PR #299, opened 2023-08-31
ROAD-536 Add VersionJump model and admin integration

arielj:
### Jira Ticket 
https://ombulabs.atlassian.net/browse/ROAD-536

### Motivation / Context

This issue https://github.com/fastruby/points/issues/296
    
### QA / Testing Instructions

- Create a Version Jump object in the admin panel
- Then it can be asigned to projects (new or existing)
- Then, for each version jump, the admin panel will show all the projects associated

### Screenshots:

image
image
image

___


app/models/project.rb
```diff
@@ -52,6 +53,10 @@ def archived?
     status == "archived"
   end
 
+  def locked?
+    locked_at.present?
+  end
```
arielj, line 58: I can revert this change, but was really simple to add (I updated some code that was using `project.locekd_at.nil?` directly)

app/policies/story_policy.rb
```diff
@@ -0,0 +1,2 @@
+class StoryPolicy < ApplicationPolicy
+end
```
arielj, line 2: I had to create all the policies to be able to use this Pundit feature in the admin panel

app/views/madmin/application/index.html.erb
```diff
@@ -11,6 +11,12 @@
         </svg>
       <% end %>
     </form>
+
+    <% if policy(resource).create? %>
+      <%= link_to resource.new_path, class: "bg-white hover:bg-gray-100 text-gray-800 font-semibold py-2 px-4 border border-gray-400 rounded shadow" do %>
+        New <span class="hidden md:inline"><%= resource.friendly_name %></span>
+      <% end %>
+    <% end %>
```
arielj, line 19: I took this back from the original gem

torresga: What's the original gem? madmin?

torresga:
@arielj One thing I noticed when testing the admin panel is that there is no way to prevent admins from creating the same version jump multiple times. For example, an admin can create multiple Rails projects that have an initial version of 6.0 and a target version of 6.1. There's also nothing stopping admins from creating a version jump that has more than one jump (think a version jump from 6.0 to 7.1 for example) This might lead to users being confused and possibly disjointed data. I don't know whether we should explore this further, or for our purposes, we can trust our users to not make mistakes. What do you think?

arielj:
> @arielj One thing I noticed when testing the admin panel is that there is no way to prevent admins from creating the same version jump multiple times. For example, an admin can create multiple Rails projects that have an initial version of 6.0 and a target version of 6.1. There's also nothing stopping admins from creating a version jump that has more than one jump (think a version jump from 6.0 to 7.1 for example) This might lead to users being confused and possibly disjointed data. I don't know whether we should explore this further, or for our purposes, we can trust our users to not make mistakes. What do you think?

@torresga I'll add a validation to prevent adding the same jump twice

About a jump of more than 1 version I think it's fine? at least I did that intentionally: I was thinking of using, for example, `Rails / 4.2 - 6.0` for the parent project, and then `Rails / 4.2 - 5.0` `Rails / 5.0 - 5.1`, `Rails / 5.1 - 5.2` and `Rails / 5.2 - 6.0` for each of the sub-projects.

I guess we'll need ernesto to define this better.

app/madmin/resources/version_jump_resource.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  # Associations
+  attribute :projects, form: false
+
+  # Uncomment this to customize the display name of records in the admin area.
+  def self.display_name(record)
+    record.to_label
+  end
+
+  # Uncomment this to customize the default sort column and direction.
```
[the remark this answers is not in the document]
arielj: This was already there commented out in the other resources. Not sure why that was not removed before, so I'd say to note remove it just for this resource (maybe some low-priority story can be about cleaning up these comments, I don't want to change files that I don't need to change for a comment)

app/views/madmin/application/index.html.erb
arielj: yes, sorry, from this file of madmin https://github.com/excid3/madmin/blob/14599a85f0fb687d6ad9bf30ee38db9f8b1abe3c/app/views/madmin/application/index.html.erb#L15

hmdros:
I would leave a small suggestion to add a badge, maybe a small one inside the box, indicating the [subproject](https://github.com/fastruby/points/pull/299/files#diff-31a7f588fa0b08d2f2cc6ea6c34a6b25125b961b0e9c63bfc77536482878d157R23) version jump:

Captura de tela 2023-09-04 094608
Captura de tela 2023-09-04 093544

arielj:
> I would leave a small suggestion to add a badge, maybe a small one inside the box, indicating the [subproject]

I did that initially, but then I realized this is how we name subprojects and it was redundant to have the same information in the name of the subproject and the label 
image

torresga (approved):

etagwerker:
> I guess we'll need ernesto to define this better.

@torresga @arielj We can trust that admins will know enough to not create jumps that span multiple versions, so I don't think this is a blocker.

arielj:
Pushed 2 updates:
- version jump column/field for projects in admin panel
image
image

And CSS issue fix:
image

etagwerker:
@arielj Looks good, thank you!

PR #302, opened 2023-09-05
ROAD-539 / Fix debounce of inputs' preview

### LINKED ISSUE #250: Story description preview fires too often
**Branch/Commit:**

Main branch.

**Expected behavior:**

When typing a description, we should wait a moment to give time to the user to stop typing before doing the request to the server to process the preview.

**Actual behavior:**

It's happening after every change, so typing many change starts to stack and acts weirdly.

**Steps to reproduce:**

How do I achieve this behavior? Use the following format to provide a step-by-step guide:

1. Go to edit/create a story
2. Start typing a description without stopping
3. the preview will be slowly updating for a while after finishing typing

arielj:
### Jira Ticket 

https://ombulabs.atlassian.net/browse/ROAD-539

### Motivation / Context

This PR fixes #250 

We have code to debounce the preview, but the code was not really doing a debounce, it was just delaying the requests but still doing all of them (and even doing extra requests since it was triggering the preview of the 2 inputs for any input change (even the title changing)).

This PR moves some things around to keep the debounceTimer in the right scope and to only update the preview of the text area that changed.
    
### QA / Testing Instructions

- Open a story edit/create form
- Open dev tools
- Start typing in the different inputs
- There should be only a few preview requests (to `/stories/render_markdown`) a moment after we stop typing instead of a request for each keystroke

___

### DIFF
--- a/app/assets/javascripts/stories.js
+++ b/app/assets/javascripts/stories.js
@@ -1,25 +1,28 @@
-document.addEventListener("input", (e) => {
-  const updateMarkdown = () => {
-    document.querySelectorAll("[data-has-preview]").forEach((element) => {
-      const form = element.closest("form");
-      const preview = form.querySelector(
-        "." + element.dataset.previewTarget + " .content"
-      );
-      if (preview) {
-        Rails.ajax({
-          type: "POST",
-          url: "/stories/render_markdown",
-          data: `markdown=${encodeURIComponent(element.value)}`,
-          dataType: "text",
-          success: (response) => {
-            preview.innerHTML = response;
-          },
-        });
-      }
-    });
-  };
+document.addEventListener("DOMContentLoaded", () => {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/stories.js
```diff
@@ -1,25 +1,28 @@
-document.addEventListener("input", (e) => {
```
arielj, line 1: this was capturing any input event, even for the title

app/assets/javascripts/stories.js
```diff
@@ -1,25 +1,28 @@
...
-  const updateMarkdown = () => {
-    document.querySelectorAll("[data-has-preview]").forEach((element) => {
```
arielj, line 3: this was doing a request for each element with preview instead of the current one that changed

Issue #195, opened 2022-01-21
Add comments to stories


arielj:
https://ombulabs.atlassian.net/browse/DT-254

**Branch/Commit:**

Main Branch

**Describe the feature:**

People working on a project would be able to add comments to each story. Comments will work as communication between the people working there but won't show up in the action plan.

**Problem:**

Discussion about stories in a project happens outside points (slack for example). This creates a problem where the story and the discussion are in 2 different places and it's hard to have visibility on both things.

**Mockups:**

When showing a story we should show a list of comments witht username and timestamp. We should show a form to create comments too.

Maybe an edit option would be useful.

We can try Rails' ActionText feature to support WYSIWYG.

**Resources:**

https://edgeguides.rubyonrails.org/action_text_overview.html

**I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)**

PR #304, opened 2023-09-11
ROAD-546: Add comments to stories

torresga:
### Jira Ticket 
[Jira ticket associated with this PR](https://ombulabs.atlassian.net/browse/ROAD-546)

### Motivation / Context
Context can be found in this issue: https://github.com/fastruby/points/issues/195

Basically, this PR adds comments to Stories in Points. This is to help keep information together that usually ends up in other channels (such as Slack). The comments are also exportable. The comments don't end up in the action plan.
    
### Screenshots:

Screen Shot 2023-09-15 at 9 32 43 AM

___

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -10,40 +10,30 @@ class CommentsController < ApplicationController
     @comment = current_user.comments.build(story: @story)
     @comment.attributes = comment_params
     saved = @comment.save
-    respond_to do |format|
-      format.html do
-        if saved
-          flash[:success] = "Comment created!"
-        else
-          flash[:error] = @comment.errors.full_messages
-        end
-
-        redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
-      end
+    if saved
+      flash[:success] = "Comment created!"
+    else
+      flash[:error] = @comment.errors.full_messages
     end
+
+    redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
   end
 
   def update
     updated = @comment.update(comment_params)
-    respond_to do |format|
-      format.html do
-        if updated
-          flash[:success] = "Comment updated!"
-          redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
-        else
-          flash[:error] = @comment.errors.full_messages
-          render :edit
-        end
-      end
+    if updated
+      flash[:success] = "Comment updated!"
+      redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
+    else
+      flash[:error] = @comment.errors.full_messages
+      render :edit
     end
   end
 
   def destroy
     @comment.destroy
-    respond_to do |format|
-      flash[:success] = "Comment deleted!"
-      format.html { redirect_to project_story_path(@comment.story.project_id, @comment.story_id) }
-    end
+    flash[:success] = "Comment deleted!"
+    redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
   end
 
   private
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -9,7 +9,6 @@ class CommentsController < ApplicationController
   def create
     @comment = current_user.comments.build(story: @story)
     @comment.attributes = comment_params
-    @comment.user_id = current_user.id
     saved = @comment.save
     respond_to do |format|
       format.html do
@@ -21,7 +20,6 @@ class CommentsController < ApplicationController
 
         redirect_to project_story_path(@comment.story.project_id, @comment.story_id)
       end
-      format.js
     end
   end
 
@@ -37,15 +35,14 @@ class CommentsController < ApplicationController
           render :edit
         end
       end
-      format.js
     end
   end
 
   def destroy
     @comment.destroy
     respond_to do |format|
-      format.html { redirect_to project_story_path(@comment.story.project_id, @comment.story_id), notice: "comment was successfully destroyed." }
-      format.json { head :no_content }
+        flash[:success] = "Comment deleted!"
+        format.html { redirect_to project_story_path(@comment.story.project_id, @comment.story_id) }
     end
   end
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,0 +1,69 @@
+class CommentsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :load_story_and_project
+  before_action :find_comment, only: [:edit, :update, :destroy]
+
+  def edit
+  end
+
+  def create
+    @comment = current_user.comments.build(story: @story)
+    @comment.attributes = comment_params
+    @comment.user_id = current_user.id
+    saved = @comment.save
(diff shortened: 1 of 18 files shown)


arielj (changes requested):
hey @torresga @hmdros, I added some comments related to the 3 stories

app/controllers/stories_controller.rb
```diff
@@ -86,9 +88,22 @@ def import
...
+          comments = {}
+          story.comments.each do |c|
+            comments["comment_#{c.id}"] = c.body
+            comments_headers.append("comment_#{c.id}")
+          end
+          csv << CSV_HEADERS + comments_headers
+          csv << story.attributes.merge(comments).slice(*CSV_HEADERS + comments_headers)
+        else
+          csv << CSV_HEADERS
+          @project.stories.by_position.each do |story|
```
arielj, line 103: this is looping through the stories inside the other loop in line 90 when comments are not checked, this is rendering all stories multiple times

image

if we have 10 stories it will render the 10 stories 10 times

arielj (changes requested):
Added more comments, specially on the export that is currently broken

app/views/comments/_comment.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="comment-card">
+  <p class="bold"><%= comment.user.name %>: <%= markdown(comment.body) %> <%= comment.created_at %><p>
+  <% if current_user == comment.user %>
+    <%= link_to "Edit Comment", edit_project_story_comment_path(project.id, story.id, comment.id), class: "link-blue" %> |
+    <%= link_to "Delete", project_story_comment_path(project.id, story.id, comment.id), method: :delete, data: { confirm: "Are you sure?", comment_id: comment.id }, remote: true , title: "Delete" %>
```
arielj, line 5:
```suggestion
    <%= link_to "Delete", project_story_comment_path(project.id, story.id, comment.id), method: :delete, data: { confirm: "Are you sure?", comment_id: comment.id }, title: "Delete" %>
```

we are not responding with a JS view and we are redirecting back, we don't need this to be an ajax request

app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= form_with model: [project, story, comment] do |form| %>
+  <%= form.text_area :body, rows: 6 %>
+  <%= form.submit class: "button green"%>
+  <%= link_to "Back", project_story_path(comment.story.project_id, comment.story_id), id:"back", class: "button" if action_name == "edit" %>
```
arielj, line 4:
```suggestion
  <%= link_to "Back", project_story_path(project, story), id: "back", class: "button" if action_name == "edit" %>
```

I think it looks cleaner to use the variables we already have for the path

also, if the update fails, the action_name is "update" and not "edit", so this button disappears

image

app/controllers/stories_controller.rb
```diff
@@ -85,12 +87,25 @@ def import
...
   def export
-    csv = CSV.generate(headers: true) { |csv|
-      csv << CSV_HEADERS
-      @project.stories.by_position.each do |story|
-        csv << story.attributes.slice(*CSV_HEADERS)
+    csv = if params[:export_with_comments] == "1"
+      CSV.generate(headers: true) do |csv|
+        csv << CSV_HEADERS.append("comment")
+        comments = []
+        @project.stories.includes(:comments).by_position.each do |story|
```
arielj, line 94: there's an issue here, every new story in the loop export all the comments of the previous stories too

Note the second story has the comment for the first story too
image

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    flash[:error] = "Comment not found"
+    redirect_to project_story_path(params[:project_id], params[:story_id])
+  end
+
+  def load_story_and_project
+    @project = Project.find(params[:project_id])
+    @story = Story.find(params[:story_id])
+  rescue ActiveRecord::RecordNotFound
+    flash[:error] = "Project or Story not found"
+    redirect_to projects_path
```
arielj, line 53: I wonder if we should handle the 2 possible exceptions individually, so in the cases where the project exists and the story doesn't we can redirect to the project instead of redirecting to the index

arielj (changes requested):
I left some comments, some are just suggestions, but the export with comments is still not correct

app/views/comments/_comment.html.erb
hmdros: What, I thought I removed this before 🤔

app/controllers/comments_controller.rb
torresga: I agree, but should we do that in this ticket or create another issue?

app/controllers/stories_controller.rb
```diff
@@ -85,12 +87,25 @@ def import
   end
 
   def export
-    csv = CSV.generate(headers: true) { |csv|
-      csv << CSV_HEADERS
-      @project.stories.by_position.each do |story|
-        csv << story.attributes.slice(*CSV_HEADERS)
+    csv = if params[:export_with_comments] == "1"
+      CSV.generate(headers: true) do |csv|
+        csv << CSV_HEADERS.append("comment")
```
arielj, line 92: this is creating a problem, we should never modify constants

every time an export is done with comments, a new `comment` column is added to the constant, so we keep adding columns until the server is restarted

image

Then, because the constant is changed, exporting without comments also creates `comment` columns

arielj (changes requested):
I added 2 comments that are not blockers

I did find another error with the exports

arielj (approved):

mateusdeap (approved):
I think we need to make sure that importing the generated CSV works. [As per Atlassian's docs](https://support.atlassian.com/jira-cloud-administration/docs/import-data-from-a-csv-file/), I believe that the comments will need to be quoted.

Then again, I'm assuming this has worked so far, so maybe nothing else is needed? Their documentation isn't super detailed as it should be, honestly. They mention how to format _a_ comment and say that we can just put multiple comments in separate columns, but don't specify if each comment needs it's own header or if it will just assume that subsequent values are comments.

I believe we need @ABizzinotto to QA this last bit of the feature.

I'm approving the PR, but I believe we need to address the above before merging. Unless someone has tried it already

arielj:
> I think we need to make sure that importing the generated CSV works. [As per Atlassian's docs](https://support.atlassian.com/jira-cloud-administration/docs/import-data-from-a-csv-file/), I believe that the comments will need to be quoted.
> 
> Then again, I'm assuming this has worked so far, so maybe nothing else is needed? Their documentation isn't super detailed as it should be, honestly. They mention how to format _a_ comment and say that we can just put multiple comments in separate columns, but don't specify if each comment needs it's own header or if it will just assume that subsequent values are comments.
> 
> I believe we need @ABizzinotto to QA this last bit of the feature.
> 
> I'm approving the PR, but I believe we need to address the above before merging. Unless someone has tried it already

The CSV export compatible with JIRA is in a different story https://ombulabs.atlassian.net/browse/VLT-5 that was not worked on yet

arielj (approved):

PR #309, opened 2023-09-28
[ROAD 534] Add state to stories

### LINKED ISSUE #196: Add a `state` attribute to stories
https://ombulabs.atlassian.net/browse/DT-253

**Branch/Commit:**

Main branch

**Describe the feature:**

We can have different states for stories: pending / approved / rejected.

A story will be created as pending and another person working on that project should approve / reject the story. Kinda like a PR review for example but really simplified.

Only `approved` stories would show up in the action plan, `rejected/pending` stories won't.

**Problem:**

Currently, when it's decided that a stories won't apply to a project, the only option to remove it from the action plan is to delete it. It's a problem because it hides information to other people working on that project, they may not know if that was discussed or not and may add the story again.

Also, it will induce other people workin on the project to review the stories and approve them so they show up in the action plan..

mateusdeap:
### Jira Ticket 
[ROAD-534](https://ombulabs.atlassian.net/browse/ROAD-534)

### Motivation / Context
Fixes #196

This PR adds a new attribute to the story model, called `approved`, which is a boolean and depending on wether it is `true`, `false` or `nil` it represents 3 possible states to a story: Approved, Rejected or Pending.

By default, all stories are created in the Pending state.

It is possible to define this state during story creation and edition.

Additionally, in the story show page, one can use a dropdown to change the state.

Finally, added labels in the project show page so we know at a glance which stories have been approved or not.
    
### QA / Testing Instructions
Run the application via `rails s` or docker and visit any project page containing stories in the stories list. You should see they are all pending.

To change the state, click on any story and then click on the dropdown button near the story title and select one of the states in the dropdown.

Additionally, you can click to edit the story and select the state from a select box or, in the project's show page, if you click to add a new story, you should be able to select it's state via a select box as well.

### Screenshots:
image
image
image
image
image
image

___

[ROAD-534]: https://ombulabs.atlassian.net/browse/ROAD-534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/migrate/20230829001347_add_approved_to_stories.rb
```diff
@@ -0,0 +1,5 @@
+class AddApprovedToStories < ActiveRecord::Migration[7.0]
+  def change
+    add_column :stories, :approved, :boolean, default: nil
```
arielj, line 3: I feel it could be more maintainable and semantic to have a `status` or `state` integer column and then use Rails Enum to handle pending/approved/rejected states.

That would also add some methods to the model like `story.approved?` or`story.pending?` or `story.approve!` to be more declarative and less imperative

app/helpers/stories_helper.rb
```diff
@@ -1,2 +1,28 @@
 module StoriesHelper
+  def status_label(story)
+    if story.pending?
+      return "<span class='story-status-badge'>pending</span>".html_safe
+    end
+
+    story.approved ? "<span class='story-status-badge approved'>approved</span>".html_safe : "<span class='story-status-badge rejected'>rejected</span>".html_safe
+  end
+
+  def status_string(story)
```
arielj, line 10: if we use an enum column for example, we wouldn't need this helper method

app/helpers/stories_helper.rb
```diff
@@ -1,2 +1,28 @@
 module StoriesHelper
+  def status_label(story)
```
arielj, line 2: if we use an enum column, we could write this as:

```
def status_label(story)
  "<span class='story-status-badge #{story.state}'>#{story.state}</span>".html_safe
end

File: app/views/projects/show.html.erb [hunk 4]
```diff
@@ -14,6 +14,7 @@
         <thead class="table-header fixed-header">
           <tr class="project-table__row project-table__row--header">
             <th class="project-table__cell">Story Title</th>
+            <th class="project-table__cell">Status</th>
```
arielj, line 17: we need to add one td for the totals at the end of the table to match the columns
image

app/views/projects/show.html.erb
```diff
@@ -25,12 +26,13 @@
         <tbody id="stories" data-url="<%= sort_stories_project_path(@project.id) %>" data-unlocked="<%= is_unlocked?(@project) %>">
           <% if @stories.present? %>
             <% @stories.each do | story | %>
-              <tr class="project-table__row project-table__row--story" id="<%= dom_id(story)%>" >
+              <tr class="project-table__row project-table__row--story" id="<%= dom_id(story)%>" data-status="<%= status_string(story) %>">
```
arielj, line 29: I understand this is here for the filtering, I would remove it from this PR for now to avoid confusion

config/routes.rb
```diff
@@ -14,6 +14,7 @@
   resource :stories do
     post :bulk_destroy, to: "stories#bulk_destroy"
     post :render_markdown
+    patch :update_status
```
arielj, line 17: this would make more sense as a member route instead of a collection route, then we could do `update_status_story_path(@story, state: "approved")` instead of `update_status_path(story_id: @story.id, state: "approved")`

app/controllers/stories_controller.rb
```diff
@@ -1,8 +1,8 @@
 require "csv"
 class StoriesController < ApplicationController
   before_action :authenticate_user!
-  before_action :find_project, except: [:bulk_destroy, :render_markdown, :edit, :update, :destroy, :show, :move]
-  before_action :find_story, only: [:edit, :update, :destroy, :show, :move]
+  before_action :find_project, except: [:bulk_destroy, :render_markdown, :edit, :update, :destroy, :show, :move, :update_status]
```
arielj, line 4: we are not sending a project_id in the request for update_status, we don't need this before_action, the `find_story` callback is setting `@project` from the story

app/assets/stylesheets/project.scss
```diff
@@ -321,3 +321,9 @@
     }
   }
 }
+
+.story-title {
```
arielj, line 325: I'd move this CSS to `stories.scss` instead of `project.scss`

config/routes.rb
arielj: I'm also wondering if something like `approve` and `reject` endpoints would be better, so we could say `approve_story_path(@story)`

db/migrate/20230829001347_add_approved_to_stories.rb
mateusdeap: Agreed

mateusdeap:
@arielj Addressed your comments

spec/rails_helper.rb
```diff
@@ -15,7 +15,7 @@
   browser_options.args << "-headless"
   Capybara::Selenium::Driver.new(app, browser: :firefox, options: browser_options)
 end
-Capybara.javascript_driver = :headless_firefox
+Capybara.javascript_driver = :selenium
```
arielj, line 18: let's revert this to headless_firefox, ir's creating issues in CI

config/routes.rb
```diff
@@ -11,6 +11,10 @@
   get "home/index"
   get "reports/index"
 
+  patch "stories/:id/approve", to: "stories#approve", as: "approve_story"
+  patch "stories/:id/reject", to: "stories#reject", as: "reject_story"
+  patch "stories/:id/pending", to: "stories#pending", as: "pending_story"
```
arielj, line 16: this could be refactored to let rails add most of the boilerplate https://guides.rubyonrails.org/routing.html#adding-member-routes

moving this to the next block like:

```rb
resource :stories do
  member do
    patch :approve
    patch :reject
    patch :pending
  end
  ...
  ...
```

I think it creates the same routes

arielj (approved):
I added one suggestion and I understand the CI tests are only failing for a configuration issue.

I'm approving this so it's ready for QA apart from those 2 comments

spec/rails_helper.rb
mateusdeap: Damn, commited by accident

arielj:
After merging, we have to run `rails db:migrate` in heroku's console.

app/helpers/stories_helper.rb
```diff
@@ -1,2 +1,13 @@
...
+    "<span class='story-status-badge #{story.status}'>#{story.status}</span>".html_safe
+  end
+
+  def status_color(story)
+    return "green" if @story.approved?
+
+    return "magenta" if @story.rejected?
+
+    return "orange" if @story.pending?
+  end
```
JuanVqz, line 12: If Rubocop is not complaining it's okay but I'm sure it will complain in the future because of the last explicit return.
Also if `pending` is the default we can remove the `if`.


```suggestion
  def status_color(story)
    return "green" if @story.approved?
    return "magenta" if @story.rejected?

    "orange"
  end
```

app/views/stories/_form.html.erb
```diff
@@ -13,6 +13,11 @@
 
   <%= f.hidden_field :project_id, value: params[:project_id] %>
 
+  <div class="field story_status">
+    <%= f.label :status, "Status" %>
+    <%= f.select :status, options_for_select({ "Pending" => "pending", "Approved" => "approved", "Rejected" => "rejected" }, selected: @story.status), class: "project-story-approved" %>
+  </div>
```
JuanVqz, line 19: With the new status field, this seems a little bit moved, I expected the preview to be aligned with the description field at least as it was before.

The first thing I want to write is the story's title. can we move the status field after the extra info?

All of this is purely personal opinion.

image

spec/features/story_show_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      status_button = find(".story-title").find(".dropdown-wrapper").find("button")
+
+      status_button.click
+      click_button("Approve")
+
+      status_button = find(".story-title").find(".dropdown-wrapper").find("button")
+
+      expect(status_button).to have_text("approved")
+
+      click_button "Save changes"
```
JuanVqz, line 41: The status got updated in the modal even without pressing the "save changes" but on the `show` page it is still the previous status. This might be a little bit weird but okay going with that.

Peek 2023-10-09 14-23

app/controllers/stories_controller.rb
```diff
@@ -127,6 +127,27 @@ def move
...
+    end
+  end
+
+  def pending
+    @story.pending!
+    respond_to do |format|
+      format.js { render "shared/update_status" }
+    end
+  end
+
```
JuanVqz, line 150: This looks weird to me because we will need to create a new endpoint if in the future we decide to add a new story's status.

what about doing something like so:

```ruby

def update_status
  @story.update(status: params[:status]) # pass a number
  # or
  @story.try(params[:status]) # where the param is a string like `pending!`, `rejected!`, etc. 
    
  respond_to do |format|
    format.js { render "shared/update_status" }
  end
end
```

```ruby
@story.update(status: params[:status]) 
irb(main):003:0> story.update(status: 0)
  TRANSACTION (0.1ms)  BEGIN
  Project Load (0.2ms)  SELECT "projects".* FROM "projects" WHERE "projects"."id" = $1 LIMIT $2  [["id", 4], ["LIMIT", 1]]
  TRANSACTION (0.2ms)  COMMIT
=> true
irb(main):004:0> story.pending?
=> true
irb(main):005:0> story.rejected?
=> false
```

```ruby
@story.try(params[:status]) 
irb(main):006:0> story.try("approved!")
  TRANSACTION (0.1ms)  BEGIN
  Story Update (0.3ms)  UPDATE "stories" SET "updated_at" = $1, "status" = $2 WHERE "stories"."id" = $3  [["updated_at", "2023-10-09 20:38:42.837573"], ["status", 1], ["id", 8]]
  TRANSACTION (0.4ms)  COMMIT
=> true
irb(main):007:0> story.pending?
=> false
irb(main):008:0> story.approved?
=> true
```

app/controllers/stories_controller.rb
```diff
@@ -143,7 +164,7 @@ def validate_url_product_id
   end
 
   def stories_params
-    params.require(:story).permit(:title, :description, :extra_info, :project_id)
+    params.require(:story).permit(:title, :description, :extra_info, :project_id, :approved)
```
JuanVqz, line 167: shouldn't this be status?

```suggestion
    params.require(:story).permit(:title, :description, :extra_info, :project_id, :status)
```

JuanVqz (commented):
Let me know what do you think

app/controllers/stories_controller.rb
arielj: this was the original and I asked to replace that with individual actions per status lol

I don't think it hurts to have new endpoints, it improves the readability (it's more clear to read `@story.approve!` than `@story.update(status: params[:status])`)

JuanVqz: Okay, that sounds good to me.

app/views/stories/_form.html.erb
mateusdeap: True. Will fix

spec/features/story_show_spec.rb
mateusdeap: True. I'll change the spec and fix up the JS

app/views/stories/_form.html.erb
mateusdeap: So, I moved the status selector, but I noticed that the description preview is actually in the right place, according to the [scss](https://github.com/fastruby/points/blob/main/app/assets/stylesheets/stories.scss#L95-L97) and the app version currently in production.

app/helpers/stories_helper.rb
```diff
@@ -1,2 +1,18 @@
 module StoriesHelper
+  def status_label(story)
+    "<span class='story-status-badge #{story.status}'>#{story.status}</span>".html_safe
```
JuanVqz, line 3: I'd like to see the status with upper case in the first letter as we saw it in the select field.

JuanVqz (approved):
I left a comment but I'll approve it since I've been so annoying. @mateusdeap Thanks!

FionaDL:
@mateusdeap feel free to merge after addressing Juan's comment (if you feel it's worth it), but don't forget to run the migrations in heroku console :)

app/helpers/stories_helper.rb
mateusdeap: I think I'm going to skip this one since I think labels are better in lowercase but, mainly, because when I started to change this, I noticed I'd have to edit quite a few specs and a bit of JS that was swapping css classes based on the text content of the label.

Maybe add this as a bug or improvement story for later on?

Issue #258, opened 2022-06-01
Prevent accidental deletion of projects like Github does


arielj:
**Branch/Commit:**

Main branch.

**Describe the feature:**

It would be nice for points to require the user to type the name of the project they want to delete instead of just asking to confirm/cancel, like Github does.

**Problem:**

Currently, the confirmation dialog is really generic and somebody can miss the text o even miss the cancel button and confirm deletion of a project.

Enforcing that the user types the project name is a good an simple measure.

**Mockups:**

This is what github shows when trying to delete a repository:
![image](https://user-images.githubusercontent.com/188464/171431360-83d243a2-c7f3-4419-8cb7-4706c1a603b0.png)

**I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)**

ABizzinotto:
As we work towards leveraging Points data more and more, preventing accidential deletion helps prevent downstream failures. Raising the priority.

PR #310, opened 2023-10-06
[MKT-575] Prevent accidental deletion in points

JuanVqz:
### Jira Ticket
https://ombulabs.atlassian.net/browse/MKT-575

Fixes https://github.com/fastruby/points/issues/258

### Motivation / Context

Prevent accidental deletion in points

### QA / Testing Instructions

Delete a project

### Screenshots:
Peek 2023-10-06 10-31

___

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,6 +1,6 @@
 class ProjectsController < ApplicationController
   before_action :authenticate_user!
-  before_action :find_project, only: [:show, :edit, :update, :sort, :sort_stories, :destroy, :new_sub_project, :toggle_archive, :toggle_locked]
+  before_action :find_project, only: [:show, :edit, :update, :sort, :sort_stories, :destroy, :new_sub_project, :toggle_archive, :toggle_locked, :open_delete_modal]
   before_action :ensure_unarchived!, only: [:edit, :new_sub_project, :update]
 
   def index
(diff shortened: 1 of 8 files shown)


app/controllers/projects_controller.rb
```diff
@@ -70,9 +70,14 @@ def create
   end
 
   def destroy
-    @project.destroy
     respond_to do |format|
-      format.html { redirect_to projects_path, notice: "Project was successfully destroyed." }
+      if @project.title.strip.eql?(params.dig(:project, :title)&.strip)
```
rishijain, line 74: Hey @JuanVqz , Out of curiosity, why are we stripping here? should it not be exact match?

spec/controllers/projects_controller_spec.rb
```diff
@@ -124,9 +124,29 @@
...
+      expect {
+        delete :destroy, params: {id: project.id, project: {title: "foo bar  "}}
+      }.to change(Project, :count).by(-1)
+    end
+
+    it "does not delete the project" do
+      expect {
+        delete :destroy, params: {id: project.id}
+      }.not_to change(Project, :count)
+    end
```
rishijain, line 149: Hey @JuanVqz , can we also add a spec for testing that it does not delete when we pass the title of the project but it is not the right title?

rishijain: I have added the spec since it was not something of a big change.

app/controllers/projects_controller.rb
JuanVqz: I noticed that sometimes when editing the project title, it left a space at the end of the title. it might be confusing if you don't realize that when deleting a project. 
Of course, we could normalize the title when saving it (strip it) and remove it here.

what do you think?

spec/controllers/projects_controller_spec.rb
JuanVqz: Thanks! I've added that test as feature :+1: 
https://github.com/fastruby/points/pull/310/files#diff-b4cf077b77d7765559bc95291a6e424d712c59271a5d6bcc7eb79b2cd3726e87R59

app/controllers/projects_controller.rb
rishijain: I do not have a strong reason as to why choose 1 approach over the other. So I would say, lets keep what you have already implemented. Thanks for clarifying.

rishijain:
The PR looks good from code review perspective.

FionaDL:
QA looks good!

Issue #318, opened 2024-01-08
Exception in admin panel (admin/stories#show)


etagwerker:
**Before we start...:**

- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)

**Branch/Commit:**

`main` branch

**Expected behavior:**

I would expect https://points.ombulabs.com/madmin/stories/9165 (madmin/stories#show) to show me a story.

**Actual behavior:**

Right now it is producing a 500 error page.

**Steps to reproduce:**

1. Sign in as an admin
2. Go to /madmin
3. Go to any stories
4. Click on the story id

**Context and environment:**

This is currently happening in production. 

**Screenshots and Videos**

![Screenshot 2024-01-08 at 10 46 06 AM](https://github.com/fastruby/points/assets/17584/5ed14ae7-bbb2-42ec-ab17-6771d79b8de6)

**Logs**

```
2024-01-08T15:13:57.611572+00:00 app[web.1]: I, [2024-01-08T15:13:57.611529 #2]  INFO -- : [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd] Completed 500 Internal Server Error in 18ms (ActiveRecord: 5.4ms | Allocations: 5256)
2024-01-08T15:13:57.613090+00:00 app[web.1]: E, [2024-01-08T15:13:57.613017 #2] ERROR -- : [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd]
2024-01-08T15:13:57.613091+00:00 app[web.1]: [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd] ActionView::Template::Error (undefined method `truncate' for nil:NilClass):
2024-01-08T15:13:57.613091+00:00 app[web.1]: [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd]     1: <% field.value(record).each do |object| %>
2024-01-08T15:13:57.613092+00:00 app[web.1]: [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd]     2:   <div>
2024-01-08T15:13:57.613092+00:00 app[web.1]: [e23d9c62-218b-46ba-a2d6-5bc6c0d5eabd]     3:     <%= link_to Madmin.resource_for(object).display_name(object), Madmin.resource_for(object).show_path(object), class: "text-indigo-500 underline" %>
[...]

PR #320, opened 2024-01-16
Update for missing name and add tests

FionaDL:
### Jira Ticket 
[DT-455](https://ombulabs.atlassian.net/browse/DT-455)

### Motivation / Context
#318 This fixes this issue. 
If anyone on the team didn't have a name in Github no name was being saved that user and it was causing issues if a user was attached to a story through an estimation. Updated so that any time there isn't a name we can fall back on the email so we won't have this issue anywhere again in the future. Also added a test to make sure this was working. 

    
### QA / Testing Instructions
Try adding changing a user in the test db to not have a name and make sure madmin still works. 
Follow the steps in issue #318 to make sure it is still fixed. 

### Screenshots:

___

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,4 +3,9 @@ class User < ApplicationRecord
 
   has_many :estimates
   has_many :comments
+
+  # Returns the email if the user doesn't have a name in Github
+  def name
+    read_attribute(:name) || email
+  end
 end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -29,4 +29,16 @@ RSpec.describe User, type: :model do
       end
     end
   end
+  describe "Instance Methods" do
+    describe "#name" do
+      it "returns the name a name exists" do
+        expect(subject.name).to eql subject.name
+      end
+
+      it "returns the email if the name doesn't exist" do
+        subject.name = nil
+        expect(subject.name).to eql subject.email
+      end
+    end
+  end
 end


spec/models/user_spec.rb
```diff
@@ -29,4 +29,16 @@
       end
     end
   end
+  describe "Instance Methods" do
```
JuanVqz, line 32:
```suggestion

  describe "Instance Methods" do
```

app/models/user.rb
```diff
@@ -3,4 +3,9 @@ class User < ApplicationRecord
 
   has_many :estimates
   has_many :comments
+
+  # Returns the email if the user doesn't have a name in Github
+  def name
+    read_attribute(:name) || email
```
JuanVqz, line 9: is the email required? what if the email is nil?

etagwerker: @JuanVqz Users come from GitHub and we always collect their email address. So there is no way for the email to be nil. 

@FionaDL On the other hand, this doesn't look very straightforward, why do we need the call to `read_attribute`?

I would recommend two different alternatives: 

```suggestion
    name || email
```

Or: 

```suggestion
    name.presence || email
```

FionaDL: @etagwerker I tested with your suggestions, but I think it is creating an infinite loop, I get "stack level too deep (SystemStackError)". 

I think we could do self[:name] || email if it feels more straightforward. What do you think?

etagwerker: @FionaDL Oh, I understand why that would do that.

How about this instead? 

```
  def name_or_email
    name || email
  end
```

spec/models/user_spec.rb
FionaDL: Rubocop didn't like this, so closing it.

app/models/user.rb
FionaDL: I could do that, but then we will have to be sure that name_or_email gets called everywhere that we would want to call `user.name`, right?

etagwerker: When it comes to presentation logic, I think it is usually best to have this sort of code in a view helper file. 

So maybe it would be best to move this code to the application_helper file and later change the references in the view to this new method.

So yeah, that means that all view files that call `user.name` will now have to call something like `display_name(user)`

etagwerker (changes requested):
@FionaDL I think this is very close. I think there is one last place to add this: https://github.com/fastruby/points/blob/c754fc7a5f97a305a49a3f0a4e9c40e1be986f89/app/controllers/stories_controller.rb#L96

etagwerker (approved):
Looks good, thanks!

PR #324, opened 2024-01-23
ROAD-609: Export approved stories only

fbuys:
### Jira Ticket 

See: https://ombulabs.atlassian.net/browse/ROAD-609

### Motivation / Context

We use the CSV tool to export stories to Jira for example. We only want to work on stories that we approved and therefore we only want to export stories from Points that were approved.

    
### QA / Testing Instructions

- Perform a CSV export in Points (without comments)
- Take note of the stories and their statuses
- Ensure that only the approved stories are present in the CSV file.
- Perform a CSV export in Points (with comments)
- Take note of the stories and their statuses
- Ensure that only the approved stories are present in the CSV file.

___

### DIFF
--- a/spec/controllers/stories_controller_spec.rb
+++ b/spec/controllers/stories_controller_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe StoriesController, type: :controller do
   render_views
 
   let!(:project) { FactoryBot.create(:project) }
-  let!(:story) { FactoryBot.create(:story, project: project) }
+  let!(:story) { FactoryBot.create(:story, project: project, status: :approved) }
 
   before do
     @request.env["devise.mapping"] = Devise.mappings[:user]
@@ -173,7 +173,9 @@ RSpec.describe StoriesController, type: :controller do
     end
 
     describe "#export" do
-      it "exports a CSV file" do
+      it "exports a CSV file with only approved stories" do
+        FactoryBot.create(:story, project: project, status: :rejected)
+        FactoryBot.create(:story, project: project, status: :pending)
         get :export, params: {project_id: project.id}
         expect(response).to have_http_status(:ok)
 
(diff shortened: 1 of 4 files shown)


spec/features/projects_manage_spec.rb
```diff
@@ -141,7 +141,9 @@
 
     context "import & Export" do
       before do
-        project.stories.create(title: "php upgrade", description: "quick php upgrade")
+        project.stories.create(title: "php upgrade", description: "quick php upgrade", status: :approved)
+        project.stories.create(title: "pending story", description: "pending stories don't export", status: :pending)
+        project.stories.create(title: "rejected story", description: "rejected stories don't export", status: :rejected)
```
JuanVqz, line 146: Didn't you think we should update those expectations? 

https://github.com/fastruby/points/blob/7b52e9feca45cf3832b60f7376614a5f6e39c70d/spec/features/projects_manage_spec.rb#L161

https://github.com/fastruby/points/blob/7b52e9feca45cf3832b60f7376614a5f6e39c70d/spec/features/projects_manage_spec.rb#L162

spec/controllers/stories_controller_spec.rb
```diff
@@ -186,11 +188,13 @@
       end
 
       context "with comments" do
-        it "exports a CSV file" do
+        it "exports a CSV file with approved stories only" do
```
JuanVqz, line 191:
```suggestion
        it "exports a CSV file with only approved stories" do
```

spec/controllers/stories_controller_spec.rb
```diff
@@ -173,7 +173,9 @@
     end
 
     describe "#export" do
-      it "exports a CSV file" do
+      it "exports a CSV file with approved stories only" do
```
JuanVqz, line 176:
```suggestion
      it "exports a CSV file with only approved stories" do
```

JuanVqz (commented):
I left a suggestion related to the expectations

spec/features/projects_manage_spec.rb
fbuys: Thank you for the feedback. To avoid confusion or false positives I moved the new stories to only be created inside of the `export` spec and not in the before_all block.

hmdros (approved):
QAed this, and all that I tested is working as expected!

PR #333, opened 2024-10-23
DT-458: Add ability for admins to export stories regardless of status

torresga:
### Jira Ticket
https://ombulabs.atlassian.net/browse/DT-458

### Motivation / Context
Currently, stories are only exported to CSV if they have an accepted status. This feature allows admin users to export stories regardless of their status. Non-admins should only be able to export approved stories.

I also refactored the `export` method because Code Climate was complaining.

### Additional changes from review
Follow-up improvements made while reviewing the PR:

- **Admin gate enforced server-side.** "Export all" is no longer trusted from the request params alone (the checkbox was only hidden in the view). The controller now honors `export_all` only when `current_user.admin?`, so a non-admin cannot forge `?export_all=1` to download non-approved stories. Added a spec for that forged-param case.
- **Fixed the admin spec.** It previously built `user.admin = true` without signing in as that user, so it never actually exercised admin behavior. It now signs in as an admin.
- **Simplified `export`.** Replaced the four-way conditional with two booleans (`with_comments`, `export_all`) and a single call to `generate_csv`, and moved `generate_csv` to `private`.
- **Removed an N+1.** With comments, the export now eager-loads comment authors (`includes(comments: :user)`), instead of firing one query per comment for `display_name`.
- **Uniform CSV rows.** A story's comments are now written into a single `comments` cell (newline-separated) instead of spilling into a variable number of unlabeled trailing columns, so every row has the same width.
- **Added a `status` column.** Since we can now export every status, the CSV includes each story's status so an all-statuses export is actually readable. Columns are ordered `id, position, status, title, description, comments`. Import matches columns by name, so the new column and ordering don't affect it.
- Added a spec covering an admin exporting all stories **with comments** (comments are included for non-approved stories too).

### QA / Testing Instructions
As an **admin**:
1. Go to a project with stories in different statuses.
2. Click **CSV Import/Export**.
3. Check **Export all stories** and click **Export** — the CSV includes stories of every status.
4. Also check **Export with comments** — the CSV has a `comments` column, populated for non-approved stories as well.

As a **non-admin**:
5. The **Export all stories** checkbox is not shown; export returns approved stories only.
6. Even hitting `…/stories/export?export_all=1` directly returns approved stories only.

### Screenshots:
Screenshot 2024-10-23 at 10 31 50 AM
___

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -90,7 +90,12 @@ class StoriesController < ApplicationController
     csv = if params[:export_with_comments] == "1"
       CSV.generate(headers: true) do |csv|
         csv << CSV_HEADERS + ["comment"]
-        @project.stories.includes(:comments).approved.by_position.each do |story|
+        stories = if params.include?(:export_all) && params[:export_all] == "1"
+          @project.stories.includes(:comments)
+        else
+          @project.stories.includes(:comments).approved
+        end
+        stories.by_position.each do |story|
           comments = []
           story.comments.each do |comment|
             comments << "#{display_name(comment.user)}: #{comment.body}"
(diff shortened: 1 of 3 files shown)


JuanVqz (approved):

JuanVqz:
QA Passed, but it is missing the status column; otherwise, we won't know what has been rejected, approved or pending, so, adding the column now.

Screenshot 2026-07-21 at 13 39 35

JuanVqz:
Now include the status, approved, and as soon as it's green, I'll merge it.

Screenshot 2026-07-21 at 13 53 18

Issue #347, opened 2025-04-15
When cloning a project, the stories' state is cloned too, they should be copied always as `pending`


arielj:
**Branch/Commit:**

main

**Expected behavior:**

When cloning a project, all stories should be marked as `pending`

**Actual behavior:**

If stories are in a different state, that state is cloned, that's incorrect.

**Steps to reproduce:**

How do I achieve this behavior? Use the following format to provide a step-by-step guide:

1. Step 1: create a project and approve/reject some stories
2. Step 2: clone it
3. Step 3: new project has some stories already marked as approved/rejected

**I will abide by the [code of conduct](https://github.com/fastruby/points/blob/main/CODE_OF_CONDUCT.md)**

JuanVqz:
@arielj, I'm just wondering, what if someone really wants to clone the stories statuses as well?

rishijain:
This was fixed by #349 . Closing this.

PR #349, opened 2025-04-22
Fix: clone stories excluding status

rishijain:
### Jira Ticket 
https://ombulabs.atlassian.net/browse/STJ-14

### Motivation / Context
Fix for #347 

While cloning the stories, I have excluded status of the story alongside some other attributes. Wrote specs to check the behaviour.
    
### QA / Testing Instructions
How do I achieve this behavior? Use the following format to provide a step-by-step guide:

Step 1: create a project and approve/reject some stories
Step 2: clone it
Step 3: new project has some stories, all of them marked as pending.

### Screenshots:

___


JuanVqz (approved):
As for QA, this is working!

It would be nice if the `ombu_labs-auth` were addressed on a separate ticket, but since we have limited time, please fix it here.

image

image

app/models/project.rb
```diff
@@ -74,7 +74,10 @@ def siblings
   end
 
   def clone_stories_into(clone)
-    stories.each { |story| clone.stories.create(story.dup.attributes) }
+    stories.each do |story|
+      attrs = story.dup.attributes.except("status", "id", "created_at", "updated_at")
```
JuanVqz, line 78: JFI: we do not need to add the `id`, `created_at`, `updated_at` the `dup` method removed them already

```suggestion
      attrs = story.dup.attributes.except(:status)
```

In the other hand, I was wondering if instead of `except` attributes we would want to accept only the "clonable" attributes we need, I'm thinking towards the future when someone adds more attributes those will be cloned and maybe we don't want that, makes sense?

